PR #59, opened 2020-12-07
Vocab manager

alepev:
I have the feeling this could be done better with ABC classes, but I'm unsure how to use them properly.. I see some repetition with having to pass `db` and `cdm` multiple times..


src/omop_etl_wrapper/model/orm_wrapper.py
```diff
@@ -98,95 +92,20 @@ def _collect_query_statistics_bulk_mode(session: Session,
...
-                # Skip unmapped records
-                if target_concept_id == 0:
-                    continue
-                self._stcm_lookup[source_vocabulary_id][source_code] = target_concept_id
-                session.add(self.cdm.SourceToConceptMap(**row))
-
-            transformation_metadata.end = datetime.now()
-            self.etl_stats.add_transformation(transformation_metadata)
-
+    @lru_cache(maxsize=50000)
```
Spayralbe, line 95: @alepev  I thought this would be a nice alternatively to storing the full STCM in memory. It allows you to have STCM mappings bigger than what fits into memory, but lookups already made earlier can be retrieved from cache.
The value of `maxsize` is maybe something we want to make configurable.

src/omop_etl_wrapper/_paths.py
```diff
@@ -18,5 +18,7 @@
 RESOURCES_DIR = Path('./resources')
 # Directory containing all source_to_concept_map files
 STCM_DIR = RESOURCES_DIR / 'source_to_concept'
+# File containing the versioning info of the STCM mappings
+STCM_VERSION_FILE = STCM_DIR / 'stcm_versions.tsv'
```
alepev, line 22: very explicit.. good!

src/omop_etl_wrapper/cdm/vocabularies/vocabularies.py
```diff
@@ -430,3 +432,20 @@ def definition_type_concept(cls):
...
+    __tablename__ = 'source_to_concept_map_version'
+    __table_args__ = {'schema': VOCAB_SCHEMA}
+
+    @declared_attr
+    def source_vocabulary_id(cls):
+        return Column(ForeignKey(f'{VOCAB_SCHEMA}.vocabulary.vocabulary_id'), primary_key=True)
+
+    @declared_attr
+    def stcm_version(cls):
+        return Column(String(255), nullable=False)
```
alepev, line 447: so the mapping version is specific per vocabulary, to only update a subset of the mappings rather than the whole table at once.. ok!

src/omop_etl_wrapper/model/vocab_manager.py
```diff
@@ -0,0 +1,41 @@
...
+        3. Deletes obsolete versions from the database;
+        4. Loads the new versions to the database.
+        :return: None
+        """
+        logger.info(f'Loading custom vocabulary tables: {self._load_custom_vocabs}')
+        if self._load_custom_vocabs:
+            self._custom_vocab_loader.load_custom_vocabulary_tables()
+
+    def load_stcm(self):
+        """Load STCM files into the source_to_concept_map table."""
```
alepev, line 38: There is a big difference in docstring length/specificity between `load_custom_vocabularies` and `load_stcm`; is this something we need to align in terms of style?

src/omop_etl_wrapper/model/orm_wrapper.py
alepev: Yes, provided we give some examples in the docs of suitable sizes, I wouldn't know what to choose without an initial reference :P

src/omop_etl_wrapper/model/stcm/stcm_loader.py
```diff
@@ -0,0 +1,167 @@
...
+from ...database import Database
+from ...model.etl_stats import EtlTransformation, EtlStats
+from ...util.io import is_hidden
+
+logger = logging.getLogger(__name__)
+
+_STCM_VERSION_TABLE_NAME = BaseSourceToConceptMapVersion.__tablename__
+
+
+class StcmLoader:
```
alepev, line 23: Oh, so splitting the different Vocabulary/Class/Concept methods into different classes didn't really help for this 🙈 OK..

src/omop_etl_wrapper/model/orm_wrapper.py
alepev: Is this something we should consider doing for other lookups, for example in the CodeMapper module? Or maybe other scenarios, there are lookups for specific CDM tables we might want to integrate.. Not sure where it could bring the most benefits, maybe something to discuss

src/omop_etl_wrapper/model/orm_wrapper.py
```diff
@@ -98,95 +92,20 @@ def _collect_query_statistics_bulk_mode(session: Session,
...
     def lookup_stcm(self, source_vocabulary_id: str, source_code: str) -> int:
```
alepev, line 96: FYI I renamed `code` into `source_concept_code` in the CodeMapper module, I think it's nice if we also agree on a common name across the repo since they represents the same thing (`source_code` / `source_concept_code` both fine by me).

src/omop_etl_wrapper/model/stcm/stcm_loader.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def _check_stcm_version_table_exists(self) -> None:
+        metadata = MetaData(bind=self._db.engine)
+        schema = self._db.schema_translate_map.get(VOCAB_SCHEMA)
+        try:
+            metadata.reflect(schema=schema, only=[_STCM_VERSION_TABLE_NAME])
+        except InvalidRequestError:
+            logger.error(f'{schema}.{_STCM_VERSION_TABLE_NAME} does not exist. '
+                         f'Run create_all to ensure all required tables are present.')
+            raise
```
alepev, line 95: will `raise` without args raise the error as expected? I thought you have to alias error as say `e`, then `raise e`; I'm just surprised it works this way!

src/omop_etl_wrapper/model/stcm/stcm_loader.py
```diff
@@ -0,0 +1,167 @@
...
+            self._loaded_stcm_versions = vocab_version_dict
+
+    def _get_provided_stcm_versions(self) -> None:
+        with STCM_VERSION_FILE.open('r') as version_file:
+            reader = csv.DictReader(version_file, delimiter='\t')
+            for line in reader:
+                vocab_id = line['source_vocabulary_id']
+                version = line['stcm_version']
+                if not vocab_id or not version:
+                    raise ValueError(f'{STCM_VERSION_FILE.name} may not contain empty values')
```
alepev, line 84: I think I should add these and other checks you are performing for the STCM table to the custom vocabulary tables..

src/omop_etl_wrapper/model/stcm/stcm_loader.py
```diff
@@ -0,0 +1,167 @@
...
+        self._update_stcm_version_table()
+
+    @staticmethod
+    def _get_stcm_files() -> Set[Path]:
+        files = set()
+        for f in STCM_DIR.glob('*'):
+            if is_hidden(f) or not f.is_file() or f.name == STCM_VERSION_FILE.name:
+                continue
+            files.add(f)
+        return files
```
alepev, line 68: I see some overlap in these methods in `CustomVocabLoader`:
```python
    @staticmethod
    def _get_all_custom_table_files() -> List[Path]:
        return [f for f in CUSTOM_VOCAB_DIR.glob('*') if f.is_file()
                and not is_hidden(f)]

    def _subset_custom_table_files(self, omop_table: str) -> List[Path]:
        # get custom vocab files for a specific vocabulary target table
        # based on the file name conventions (e.g. "concept")
        custom_table_files = self._get_all_custom_table_files()
        return [f for f in custom_table_files if f.stem.endswith(omop_table)]
```
shall we bother move the first part to an external module the classes can both import from? otherwise I will refactor those 2 to be a single method like yours

src/omop_etl_wrapper/model/stcm/stcm_loader.py
alepev: We'll reuse them for standard vocabulary updates 😬

src/omop_etl_wrapper/model/stcm/stcm_loader.py
```diff
@@ -0,0 +1,167 @@
...
+class StcmLoader:
+    def __init__(self, db: Database, cdm, etl_stats: EtlStats):
+        self._db = db
+        self._cdm = cdm
+        self._etl_stats = etl_stats
+        self._loaded_stcm_versions: Dict[str, str] = {}
+        self._new_stcm_versions: Dict[str, str] = {}
+
+    @property
+    @lru_cache()
```
alepev, line 32: general question: how can you force this to drop cached value and reload? or it happens automatically if the properties being used (`self._new_stcm_versions`, `self._loaded_stcm_versions`) are updated?

src/omop_etl_wrapper/model/stcm/stcm_loader.py
```diff
@@ -0,0 +1,167 @@
...
+    def _stcm_vocabs_to_update(self) -> Set[str]:
+        return {vocab_id for vocab_id, version in self._new_stcm_versions.items()
+                if version != self._loaded_stcm_versions.get(vocab_id)}
```
alepev, line 35: wouldn't you run into an issue if a vocabulary has never been used before? shouldn't you specify a default return value for the `.get()` method in case the `vocab_id` is not found?

alepev: ..or is this because you should in theory have loaded any new custom vocabulary before running `load_stcm()`?

src/omop_etl_wrapper/model/stcm/stcm_loader.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def _delete_outdated_stcm_records(self) -> None:
+        if not self._stcm_vocabs_to_update:
+            return
+        logger.info(f'Deleting STCM records for vocabulary_ids: {self._stcm_vocabs_to_update}')
+        with self._db.session_scope() as session:
+            stcm_table = self._cdm.SourceToConceptMap
+            q = session.query(stcm_table)
+            q = q.filter(stcm_table.source_vocabulary_id.in_(self._stcm_vocabs_to_update))
+            q.delete(synchronize_session=False)
```
alepev, line 121: I normally write the whole query in one go, now I'm thinking if chopping it up like this would help to get more specific error messages, since each part is on a separate line of code.. is that why you do it?

src/omop_etl_wrapper/model/stcm/stcm_loader.py
```diff
@@ -0,0 +1,167 @@
...
+            metadata.reflect(schema=schema, only=[_STCM_VERSION_TABLE_NAME])
+        except InvalidRequestError:
+            logger.error(f'{schema}.{_STCM_VERSION_TABLE_NAME} does not exist. '
+                         f'Run create_all to ensure all required tables are present.')
+            raise
+
+    def _must_be_parsed(self, stcm_file: Path) -> bool:
+        stcm_file_vocab_id = self._get_stcm_file_vocab_id(stcm_file)
+        if stcm_file_vocab_id is None:
+            return True
```
alepev, line 100: so files with names unrelated to vocabulary ids are always parsed; I guess the idea is either to have a single STCM file, or multiple ones per vocabulary, but not both right? (documentation note)

src/omop_etl_wrapper/model/stcm/stcm_loader.py
```diff
@@ -0,0 +1,167 @@
...
+        Insert STCM file into the STCM vocabulary table and add
+        contents to stcm_lookup.
+
+        :param stcm_file: Path
+            Separated values file with header matching the CDM STCM
+            columns
+        :return: None
+        """
+        logger.info(f'Loading STCM file: {stcm_file.name}')
+        transformation_metadata = EtlTransformation(name=f'load_{stcm_file.stem}')
```
alepev, line 148: I completely neglected `EtlTransformation`, maybe something I can add to the CustomVocabLoader too for nice logging

src/omop_etl_wrapper/model/stcm/stcm_loader.py
alepev: Just noticed you have this check in the body of `_load_stcm_from_file()`

alepev:
@Spayralbe done reviewing! I noticed several instances where we could align, not just docstrings, but I'm not going to touch the CustomVocabLoader for now. Let's discuss it, we could even decide to merge this before I touch that part again..

src/omop_etl_wrapper/model/orm_wrapper.py
Spayralbe: Might be we could use it there as well, so would be useful to discuss. The lookup function will have to change in that case though, as to cache results, the input arguments need to be hashable (e.g. no lists can be provided as function input).

src/omop_etl_wrapper/model/vocab_manager.py
Spayralbe: I've expanded the one for `load_stcm` now, but would still be good to align indeed.

src/omop_etl_wrapper/model/orm_wrapper.py
Spayralbe: I agree it would be good to standardize the names. Problem is the CDM is also not sure about the right name :p
The concept table has column `concept_code` whereas the STCM table has column `source_code`. I thought it would be less confusing to stick to the name in the STCM table.

src/omop_etl_wrapper/model/stcm/stcm_loader.py
Spayralbe: Using raise without an explicit exception like here only works because it's within an except clause. It will then simply raise the original exception that made it enter the clause.

src/omop_etl_wrapper/model/stcm/stcm_loader.py
Spayralbe: If the cache needs to be invalidated, you need to explicitly call it. See https://github.com/thehyve/omop-etl-wrapper/blob/master/src/omop_etl_wrapper/database/constraints/constraint_manager.py#L100

src/omop_etl_wrapper/model/stcm/stcm_loader.py
Spayralbe: I've updated the parsing of the stcm_version file, to immediately raise an exception if there is a vocabulary_id in there that is not already present in the vocabulary table.
Any records in an STCM file that have a source_vocabulary_id value that is not in the stcm_version file will be ignored. Maybe it's better to log a warning when that occurs? 🤔

src/omop_etl_wrapper/model/stcm/stcm_loader.py
Spayralbe: I'm afraid it doesn't really make a difference for the error message.
I only use this because I prefer it over using `\` for line continuation. That is also what PEP8 recommends.
Then again, the SQLAlchemy documentation contains many examples using `\` for queries, so you could argue for either way.

src/omop_etl_wrapper/model/stcm/stcm_loader.py
Spayralbe: Currently both are supported indeed. Having it one way or the other is cleaner, but might also be annoyingly restrictive :p

src/omop_etl_wrapper/model/stcm/stcm_loader.py
Spayralbe: Decided to log a warning

Issue #65, opened 2020-12-08
Import CodeMapper code


alepev:
CodeMapper allows to create a dictionary of mappings from non-standard codes to standard concept ids, for any valid non-standard OMOP vocabulary; optionally you can restrict the mapping dictionary to a subset of codes (e.g. those that are actually used in the source data) to reduce the dictionary size.

(Copy code over from repo it is currently in).

PR #66, opened 2020-12-08
Code mapper

alepev:
Solves #65 . Can be tested by applying changes in https://github.com/thehyve/ohdsi-etl-template/pull/49 (until proper test suite ready).


src/omop_etl_wrapper/model/mapping/code_mapper.py
```diff
@@ -0,0 +1,252 @@
...
+        """
+        Mainly for ICD9 and ICD10 codes that are recorded in the source without a dot.
+        :return:
+        """
+        new_mapping_dict = {}
+        for key in self.mapping_dict:
+            value = self.mapping_dict.get(key)
+            key_no_dot = key.replace('.', '')
+            new_mapping_dict[key_no_dot] = value
+        self.mapping_dict = new_mapping_dict
```
alepev, line 87: I think we should get rid of this, it's a bit too specific

src/omop_etl_wrapper/model/mapping/code_mapper.py
```diff
@@ -0,0 +1,252 @@
...
+                                self.cdm.ConceptRelationship.relationship_id == 'Maps to')) \
+                .outerjoin(target,
+                           and_(self.cdm.ConceptRelationship.concept_id_2 == target.concept_id,
+                                target.standard_concept == 'S',
+                                target.invalid_reason == None)) \
+                .filter(and_(*source_filters)) \
+                .all()
+
+        mapping_df = pd.DataFrame(records, dtype='object')
+        mapping_dict = MappingDict.from_mapping_df(mapping_df)
```
alepev, line 232: I wanted to replace this with a method that creates the `MappingDict` directly from `records` without having to pass via a `df`, it was a low prio issue that we never got around to in the other repo.. but would definitely improve performance

src/omop_etl_wrapper/model/mapping/code_mapper.py
```diff
@@ -0,0 +1,270 @@
...
+    def __init__(self, database, cdm):
+        self.db = database
+        self.cdm = cdm
+
+    def generate_code_mapping_dictionary(
+            self,
+            vocabulary_id: Union[str, List[str]],
+            restrict_to_codes: Optional[List[str]] = None,
+            invalid_reason: Optional[Union[str, List[str]]] = None,
+            standard_concept: Optional[Union[str, List[Union[str, int]]]] = None,
```
Spayralbe, line 156: Why also `int`? It's not one of the options listed in the docstring

src/omop_etl_wrapper/model/mapping/code_mapper.py
```diff
@@ -0,0 +1,270 @@
...
+        :param remove_dot_from_codes: bool, default False
+            For e.g. icd9 and icd10 the source codes do not contain the
+            dot separator
+        :return: MappingDict
+        """
+
+        # make sure restrict to codes contains unique and not null
+        # elements
+        if restrict_to_codes:
+            restrict_to_codes = list(filter(lambda x: not is_null(x), set(restrict_to_codes)))
```
Spayralbe, line 190: Could this also be done with a list comprehension, or by splitting into multiple lines? It's a bit hard to read currently.

src/omop_etl_wrapper/model/mapping/code_mapper.py
```diff
@@ -0,0 +1,270 @@
...
+
+        logger.info(f'Building mapping dictionary for vocabularies: {vocabulary_id}')
+
+        source = aliased(self.cdm.Concept)
+        target = aliased(self.cdm.Concept)
```
Spayralbe, line 195: Cool! Didn't know about that function.

src/omop_etl_wrapper/model/mapping/code_mapper.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def __init__(self, database, cdm):
+        self.db = database
+        self.cdm = cdm
+
+    def generate_code_mapping_dictionary(
+            self,
+            vocabulary_id: Union[str, List[str]],
+            restrict_to_codes: Optional[List[str]] = None,
+            invalid_reason: Optional[Union[str, List[str]]] = None,
```
Spayralbe, line 155: Could it be an option to enforce providing `Optional[List[str]]` for these types of arguments? It would also make the processing logic simpler, as a list containing only one element can still be used in an `in_` query.

src/omop_etl_wrapper/util/helper.py
```diff
@@ -3,6 +3,13 @@
 import pandas as pd
 
 
+def is_null(value) -> bool:
```
Spayralbe, line 6: Maybe rename this to `is_null_or_falsy`? It's otherwise easily confused with pandas `isnull` method.

src/omop_etl_wrapper/model/mapping/code_mapper.py
alepev: I need to change this, that `int` doesn't make any sense

src/omop_etl_wrapper/model/mapping/code_mapper.py
alepev: is there any difference in performance between using `in_` list of 1 vs `==` string? well, the effort is made already, I don't mind leaving it.. I see this sometimes in pandas methods and kind of like the option of using either type :P

src/omop_etl_wrapper/model/mapping/code_mapper.py
alepev: I'll rewrite it

src/omop_etl_wrapper/model/mapping/code_mapper.py
```diff
@@ -187,22 +201,25 @@ def generate_code_mapping_dictionary(
...
         elif type(vocabulary_id) == str:
             source_filters.append(source.vocabulary_id == vocabulary_id)
-        # invalid reason: either list, str (incl. NULL), or None (filter is not applied)
+        # invalid reason: either list, str (incl. NULL), or None (filter
+        # is not applied)
         if type(invalid_reason) == list:
             source_filters.append(source.invalid_reason.in_(invalid_reason))
         elif invalid_reason == 'NULL':
-            source_filters.append(source.invalid_reason == None)
+            source_filters.append(source.invalid_reason is None)
```
alepev, line 209: this is funny, I saw the warning but I really thought it was something specific to SQLAlchemy.. in the reference I used the code was `== None`, I should have tested this with the more pythonic comparison

src/omop_etl_wrapper/model/mapping/code_mapper.py
Spayralbe: Would be a nice improvement indeed.

src/omop_etl_wrapper/util/helper.py
alepev: 👌

src/omop_etl_wrapper/model/mapping/code_mapper.py
alepev: @Spayralbe any strong feelings about this? the way it works now is not performance-friendly because 1. first the code retrieves the mappings for, say, all ICD10 codes in the OMOP vocabularies, 2. then it removes the dot from the keys in the resulting mapping vocabulary 3. then it compares the keys to the list of provided ICD10 codes (if any), i.e. `restrict_to_codes`. if a key is not found in the list, it's dropped. 
it saves memory by returning a smaller dictionary, but it's still slow in creating it because it loads all the mapping.. ideally I would prefer a function that puts back the dot in the input ICD10 codes before they are passed to the `restrict_to_codes` argument, if that's even possible

src/omop_etl_wrapper/model/mapping/code_mapper.py
alepev: incredible, you learned something from me 😱

src/omop_etl_wrapper/model/mapping/code_mapper.py
Spayralbe: I think it's better to remove it then. If you really need to work with ICD codes without dots, you probably just want to load the vocabularies without the dots in the first place.

src/omop_etl_wrapper/model/mapping/code_mapper.py
Spayralbe: Far from the first time actually ;)

src/omop_etl_wrapper/model/mapping/code_mapper.py
Spayralbe: Sure I'm fine with leaving it. For performance it wouldn't matter though, as the resulting SQL query will be optimized to the same procedure.

alepev: okies

src/omop_etl_wrapper/model/mapping/code_mapper.py
```diff
@@ -0,0 +1,315 @@
...
+            mapping.source_standard_concept = record.source_standard_concept
+            mapping.source_invalid_reason = record.source_invalid_reason
+            mapping.target_concept_code = record.target_concept_code
+            mapping.target_concept_id = target_concept_id
+            mapping.target_concept_name = record.target_concept_name
+            mapping.target_vocabulary_id = record.target_vocabulary_id
+
+            mapping_dict[code] = mapping_dict.get(code, []) + [mapping]
+
+        mapping_dict_from_records.mapping_dict = mapping_dict
```
alepev, line 89: there it is! I had to create a NamedTuple for type checking because SQLAlchemy wouldn't let me import `sqlalchemy.util._collections.result`.. but actually this is a lot better (and now we can actually test this with a named tuple, without even querying anything)

src/omop_etl_wrapper/model/mapping/code_mapper.py
```diff
@@ -0,0 +1,315 @@
...
+
+        return mapping_dict_from_records
+
+    @classmethod
+    def from_mapping_df(cls, mapping_df: pd.DataFrame) -> MappingDict:
```
alepev, line 94: not sure what to do with this.. in theory it could be handy to create a mapping dict from file in specific situations, but I think we always go for the Usagi format anyway, or STCM, so I doubt we will ever use this..

src/omop_etl_wrapper/model/mapping/code_mapper.py
alepev: I just read about ICDX ontologies, in theory the dot is always after the 3rd character. so for example if you have `J3401` it would be very easy to transform it into `J34.01`. So a lot more efficient to add the dot to the source codes beforehand! @MaximMoinat is this an acceptable solution? shall we try this for UKB and see how the ETL speeds up?

alepev: ..also, if you are aware of any strange cases where this would not work

src/omop_etl_wrapper/model/mapping/code_mapper.py
alepev: nailed it

src/omop_etl_wrapper/model/mapping/code_mapper.py
alepev: unless.. we use adapt this to load a Usagi-format vocabulary? 💡

src/omop_etl_wrapper/model/mapping/code_mapper.py
alepev: Removed code

alepev:
@Spayralbe finished with this, next on.. the other PR

src/omop_etl_wrapper/model/mapping/code_mapper.py
```diff
@@ -0,0 +1,271 @@
...
+            mapping.target_concept_name = record.target_concept_name
+            mapping.target_vocabulary_id = record.target_vocabulary_id
+
+            mapping_dict[code] = mapping_dict.get(code, []) + [mapping]
+
+        mapping_dict_from_records.mapping_dict = mapping_dict
+
+        return mapping_dict_from_records
+
+    def lookup(self, code: str,
```
Spayralbe, line 102: Is `code` always a `source_code` here? May it's a bit more clear if named as such.

Spayralbe (approved):

PR #68, opened 2020-12-09
Enforce max-doc-length

Spayralbe:
Enforce max-doc-length=72 in flake8

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,5 @@
 [flake8]
 max-line-length = 99
+max-doc-length = 72
 ignore = E501, E203, F401
 exclude = _version.py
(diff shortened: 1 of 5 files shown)


.flake8
```diff
@@ -1,4 +1,5 @@
 [flake8]
 max-line-length = 99
+max-doc-length = 72
```
alepev, line 3: this is going to mow through my future commits like the Grim Reaper 😅 fair enough!

alepev (approved):

.flake8
Spayralbe: 😈

PR #70, opened 2020-12-11
Add source_data_folder to MainConfig

### LINKED ISSUE #64: Move source_data_folder to config
The source folder has to be set by the data provider. Would be nice to have all 'flexible' configs in one file.
`source_data_folder: './resources/synthetic_data'`

The source_config can still contain the schema of the files.

Spayralbe:
Resolves #64 
Depends on https://github.com/thehyve/ohdsi-etl-template/pull/50


src/omop_etl_wrapper/config/models/main_config.py
```diff
@@ -31,6 +30,7 @@ class _RunOptions(BaseModel):
 
 class MainConfig(BaseModel):
     database: _DataBase
+    source_data_folder: Optional[DirectoryPath]
```
alepev, line 33: would you say it is better to make it optional in case we want to set up the target database before we actually have input data (synthetic or real)? OK, I think it makes sense at the very beginning..

src/omop_etl_wrapper/wrapper.py
```diff
@@ -48,17 +48,22 @@ def __init__(self, config: MainConfig, cdm_):
...
         self.vocab_loader = VocabularyLoader(self.db, cdm_)
-        self.load_custom_vocabs: bool = \
-            not config.run_options.skip_custom_vocabulary_loading
+        self.load_custom_vocabs: bool = not config.run_options.skip_custom_vocabulary_loading
 
-    def _set_source_data(self):
+    def _set_source_data(self, source_data_path: Optional[Path]):
+        if source_data_path is None:
+            logger.info(f'No source_data_folder provided in config file, '
+                        f'assuming no source data files are present')
```
alepev, line 59: 👍

alepev (approved):

src/omop_etl_wrapper/config/models/main_config.py
Spayralbe: Although we often have the source data as files, it could be that the source data is already in a database, meaning you don't need this parameter at all. So that's why it's optional.

PR #75, opened 2020-12-16
Refactor EtlStats

Spayralbe:
Depends on https://github.com/thehyve/ohdsi-etl-template/pull/51
- Get rid of the single EtlStats instance being passed around to many different objects as part of the initialization. Instead there is now a single instance created in the etl_stats module, that is imported in other modules when needed.
- Wrapper now has a single `summarize` method that handles both the logging of the summary report and writing the report files, exact behavior depending on the contents of `MainConfig`.

### DIFF
--- a/src/omop_etl_wrapper/wrapper.py
+++ b/src/omop_etl_wrapper/wrapper.py
@@ -129,3 +129,8 @@ class Wrapper(OrmWrapper, RawSqlWrapper):
                 if not conn.dialect.has_schema(conn, schema_name):
                     logger.info(f'Creating schema: {schema_name}')
                     self.db.engine.execute(CreateSchema(schema_name))
+
+    def summarize(self) -> None:
+        etl_stats.log_summary()
+        if self._config.run_options.write_reports:
+            etl_stats.write_summary_files()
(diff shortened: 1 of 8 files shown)


src/omop_etl_wrapper/wrapper.py
```diff
@@ -38,25 +38,22 @@ def __init__(self, config: MainConfig, cdm_):
         :param cdm_: Module containing the SQLAlchemy declarative Base
             and the CDM tables.
         """
+        etl_stats.reset()
```
alepev, line 41: what would happen if you don't reset this? wouldn't a new instance be created anyway when you re-run the ETL?

src/omop_etl_wrapper/wrapper.py
```diff
@@ -132,3 +129,18 @@ def create_schemas(self) -> None:
...
+        Logs an overview of all transformations and data sources that
+        were used for the current run. If **write_reports** is set to
+        True in the main config file, two overview files will be written
+        to the logs folder with more detailed information.
+
+        :return: None
+        """
+        if self._config.run_options.write_reports:
+            etl_stats.write_summary_files()
+        etl_stats.log_summary()
```
alepev, line 146: nice and tidi(er)!

alepev (approved):
I was surprised you create an instance of `etl_stats` in the same file where you define the class, I don't think I have seen this before but it does seem to simplify a lot everywhere else in the code. Nice!

src/omop_etl_wrapper/wrapper.py
Spayralbe: Yes, it's only for the odd case in which a new Wrapper is instantiated within the same python session.

PR #79, opened 2020-12-17
Drop/add all constraints methods

### LINKED ISSUE #55: Method for dropping all constraints and indexes
Current `ConstraintManager` class does not include a method for dropping all constraints on vocabulary tables, only for all non-vocab tables.
However, to drop vocabulary constraints, you would also need to drop non-vocab constraints as they contain many FKs to the vocab tables. Therefore drop_all_constraints/add_all_constraints methods (that add/drop both vocab and non-vocab constraints and indexes) would be useful.

Spayralbe:
Closes #55 
- Add methods to the `ConstraintManager` for adding and dropping all constraints/indexes (including vocabulary tables).
- Add check constraints on vocabulary tables (were added in cdm6.0.0).
- Several small fixes/improvements in `ConstraintManager` logic

### DIFF
--- a/src/omop_etl_wrapper/cdm/vocabularies/vocabularies.py
+++ b/src/omop_etl_wrapper/cdm/vocabularies/vocabularies.py
@@ -1,6 +1,7 @@
 import datetime
 
-from sqlalchemy import Column, Date, ForeignKey, Integer, Numeric, String, Text, DateTime
+from sqlalchemy import (Column, Date, ForeignKey, Integer, Numeric, String,
+                        Text, DateTime, CheckConstraint)
 from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy.orm import relationship
 
@@ -9,7 +10,17 @@ from .._schema_placeholders import VOCAB_SCHEMA
 
 class BaseConcept:
     __tablename__ = 'concept'
-    __table_args__ = {'schema': VOCAB_SCHEMA}
+    __table_args__ = (
+        CheckConstraint("(COALESCE(invalid_reason, 'D'::character varying))::text "
+                        "= ANY ((ARRAY['D'::character varying, "
+                        "'U'::character varying])::text[])", name="chk_c_invalid_reason"),
+        CheckConstraint("(COALESCE(standard_concept, 'C'::character varying))::text "
+                        "= ANY ((ARRAY['C'::character varying, "
+                        "'S'::character varying])::text[])", name="chk_c_standard_concept"),
+        CheckConstraint("(concept_code)::text <> ''::text", name="chk_c_concept_code"),
+        CheckConstraint("(concept_name)::text <> ''::text", name="chk_c_concept_name"),
+        {'schema': VOCAB_SCHEMA},
+    )
 
     @declared_attr
     def concept_id(cls):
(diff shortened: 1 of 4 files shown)


src/omop_etl_wrapper/database/constraints/constraint_manager.py
```diff
@@ -320,6 +367,32 @@ def add_index(self, index: str) -> None:
         """
         self._add_constraint_or_index(index)
 
+    @staticmethod
+    def _get_table_objects(tables: List[Table],
```
alepev, line 371: This simplifies a lot of method-specific code above :)

tests/python/database/constraints/test_constraints.py
```diff
@@ -86,12 +86,21 @@ def test_drop_and_add_pk(cdm600_wrapper_with_tables_created: Wrapper):
 
 
 @pytest.mark.usefixtures("container", "test_db")
-def test_drop_index_raises_keyerror_if_missing(cdm600_wrapper_with_tables_created: Wrapper):
+def test_raise_keyerror_if_object_not_found(cdm600_wrapper_with_tables_created: Wrapper):
     wrapper = cdm600_wrapper_with_tables_created
 
+    # Cannot add/drop for non-existing tables
+    with pytest.raises(KeyError):
+        wrapper.db.constraint_manager.add_table_constraints('we_lay_my_love_and_i')
```
alepev, line 94: what? 😂

src/omop_etl_wrapper/cdm/vocabularies/vocabularies.py
```diff
@@ -9,7 +10,17 @@
...
 class BaseConcept:
     __tablename__ = 'concept'
-    __table_args__ = {'schema': VOCAB_SCHEMA}
+    __table_args__ = (
+        CheckConstraint("(COALESCE(invalid_reason, 'D'::character varying))::text "
+                        "= ANY ((ARRAY['D'::character varying, "
+                        "'U'::character varying])::text[])", name="chk_c_invalid_reason"),
+        CheckConstraint("(COALESCE(standard_concept, 'C'::character varying))::text "
+                        "= ANY ((ARRAY['C'::character varying, "
+                        "'S'::character varying])::text[])", name="chk_c_standard_concept"),
```
alepev, line 19: this is so ugly, but I guess we can't do much about it..

alepev: is it Postgres-specific synthax or more generally applicable?

alepev (approved):

src/omop_etl_wrapper/cdm/vocabularies/vocabularies.py
Spayralbe: Actually I screwed up here, because I took the postgres internal notation of the check constraints, instead of the less dialect specific notation from the cdm constraints file.
Now it looks slightly better :p

tests/python/database/constraints/test_constraints.py
Spayralbe: Can't write tests without making some obscure [references](https://en.wikipedia.org/wiki/The_Innocents_(1961_film)) ;)

Issue #71, opened 2020-12-15
Refactor CustomVocabLoader


alepev:
Integrate missing functionality and harmonize with `StcmLoader`:
- [x] speed up loading of custom concepts by checking if file name contains valid vocabulary prefix, as in (**NOTE: check if same as already described in #57**) : https://github.com/thehyve/omop-etl-wrapper/blob/e57db47ccfce66c09f25c5ec5d961cd067e66597/src/omop_etl_wrapper/model/stcm/stcm_loader.py#L103-L109 
- [x] add `EtlTransformation`, as in: 
   https://github.com/thehyve/omop-etl-wrapper/blob/e57db47ccfce66c09f25c5ec5d961cd067e66597/src/omop_etl_wrapper/model/stcm/stcm_loader.py#L143-L145
- [x] upfront validation of input tables (custom Vocabulary, ConceptClass)
- [x] per-row validation of input tables (custom Concept)
- [x] review docstrings style (e.g. dashes instead of numbered list, indentation, line spacing)
- [x] refactor these methods to reuse shared code?
   In `StcmLoader`: https://github.com/thehyve/omop-etl-wrapper/blob/e57db47ccfce66c09f25c5ec5d961cd067e66597/src/omop_etl_wrapper/model/stcm/stcm_loader.py#L64-L71   In `CustomVocabLoader`: https://github.com/thehyve/omop-etl-wrapper/blob/e57db47ccfce66c09f25c5ec5d961cd067e66597/src/omop_etl_wrapper/model/custom_vocab/custom_vocab_loader.py#L28-L37

Issue #53, opened 2020-11-27
Implement methods to remove obsolete custom vocabularies and classes


alepev:
Old versions of custom vocabularies and classes are dropped when a new version is found, but always assuming the `vocabulary_id` and `class_id` remain unchanged (only names and details are allowed to change). If the `id` changes, older versions would not be recognized as such and would remain as clutter in the vocabulary tables, together with the `concept_id`s referencing them.

The steps to remove this clutter (before loading the new custom vocabs/classes) would be:
- get a list of currently loaded custom vocabularies and classes (`vocabulary_concept_id == 0` and `concept_class_concept_id == 0`, while standard Athena vocabularies and classes always have a specific `concept_id`)
- find any vocabs/classes whose `id` is not in the current custom vocabularies/classes files to load
- remove all `concept_id`s associated with the obsolete vocabularies/classes
- remove the obsolete vocabularies/classes

Example of methods to implement in `VocabularyLoader`:
```python
unused_custom_vocabs = self._get_list_of_unused_custom_vocabularies()
self._drop_unused_custom_concepts(unused_custom_vocabs)
self._drop_unused_custom_vocabularies(unused_custom_vocabs)
unused_custom_classes = self._get_list_of_unused_custom_classes()
self._drop_unused_custom_classes(unused_custom_classes)
```

Additional TODOs:
- [x] instead of performing DELETE when class name changes, UPDATE name field

PR #81, opened 2020-12-22
Custom vocab refactoring

alepev:
Addresses #71 and #53

Related changes in template: https://github.com/thehyve/ohdsi-etl-template/pull/52

Main changes:
- obsolete custom vocabularies and classes are removed automatically from the database
- custom concept_class names can change independently from vocabularies (now performed with UPDATE rather than DELETE + INSERT)
- check concept table names to speed up loading (load if either no prefix, prefix unrelated to vocabulary names, or prefix matching one of the vocab_ids to load - if prefix matches valid vocab_id but not one of those to load, skip)
- quality checks for user-provided custom vocabulary, concept class, concept tables
- moved some code to `utils/io.py` and applied minor refactoring to `stcm_loader` accordingly
- fixed a problem in `database.py` introduced by CheckConstraint
- added `etl_stats` to custom vocabulary loader base classes (though it doesn't work for updates and deletions! -> REMOVE?)


src/omop_etl_wrapper/database/database.py
```diff
@@ -146,7 +146,13 @@ def get_record_targets(record_containing_object: Iterable) -> Iterable[str]:
...
         """
         for record in record_containing_object:
-            placeholder_schema = record.__table_args__.get('schema')
+            # if table constraints present (SQLAlchemy CheckConstraint),
+            # __table_args__ consists of a tuple where the dictionary
+            # containing the schema name is listed as the last element
+            if len(record.__table_args__) > 1:
+                placeholder_schema = record.__table_args__[-1].get('schema')
+            else:
+                placeholder_schema = record.__table_args__.get('schema')
```
alepev, line 155: this is a very funny thing I discovered, apparently a consequence of the new constraint changes

src/omop_etl_wrapper/model/custom_vocab/base_manager/base_concept_manager.py
```diff
@@ -18,46 +20,113 @@ def __init__(self, db: Database, cdm, custom_concept_files: List[Path]):
...
+                                             f'vocabulary_id')
+                        if not concept_class_id:
+                            raise ValueError(f'{concept_file.name} may not contain an empty '
+                                             f'concept_class_id')
+                        if not valid_start_date or not valid_end_date:
+                            raise ValueError(f'{concept_file.name} may not contain empty '
+                                             f'date fields')
+                        if not domain_id:
+                            raise ValueError(f'{concept_file.name} may not contain an empty '
+                                             f'domain_id')
```
alepev, line 96: I wonder if these even make sense, given that if any of these conditions is not met, it will fail immediately afterwards in the record creation steps...

src/omop_etl_wrapper/model/custom_vocab/custom_vocab_loader.py
```diff
@@ -26,26 +26,57 @@ def __init__(self, db: Database, cdm):
...
+        # filename is unrelated to vocab ids
+        if custom_file_vocab_id not in self.vocabs_from_disk.keys():
+            return True
+        # filename contains vocab_id that matches vocabs to update
+        if custom_file_vocab_id in self._custom_vocabs_to_update:
+            return True
+        # filename contains no prefix
+        if custom_file_vocab_id is None:
+            return True
+        return False
```
alepev, line 53: this is the same logic as in stcm_loader, except here I wrote some comments. we can add it to the other file too, or make this a common method

src/omop_etl_wrapper/model/stcm/stcm_loader.py
```diff
@@ -99,21 +95,15 @@ def _check_stcm_version_table_exists(self) -> None:
...
-        stcm_file_vocab_id = self._get_stcm_file_vocab_id(stcm_file)
-        if stcm_file_vocab_id is None:
+        stcm_file_vocab_id = get_file_prefix(stcm_file, 'stcm')
+        if stcm_file_vocab_id not in self._provided_stcm_versions:
             return True
         if stcm_file_vocab_id in self._stcm_vocabs_to_update:
             return True
+        if stcm_file_vocab_id is None:
+            return True
         return False
```
alepev, line 105: I hope I did it right

alepev:
One last note on this: I'm not making any use of methods to drop/recreate constraints at the moment; should I? and what about indexes?

src/omop_etl_wrapper/database/database.py
Spayralbe: Glad you found this! I think it can be done in a slightly simpler way by not using the `__table_args__` at all, but instead use the `__table__` variable:
```suggestion
            placeholder_schema = record.__table__.schema
```

src/omop_etl_wrapper/model/custom_vocab/base_manager/base_concept_manager.py
Spayralbe: Yes I think they can be removed. The error you get when one of these values is empty is already pretty clear.

Spayralbe:
> One last note on this: I'm not making any use of methods to drop/recreate constraints at the moment; should I? and what about indexes?

No need to. It's unlikely you save time by first dropping these. Also we cannot know beforehand whether it would save time or not, so it's best to leave making that call to the user.

Spayralbe (commented):
Nice improvements 👌 Clear variable naming and useful clarifying comments in the private methods.
Some suggestions in the comments.

src/omop_etl_wrapper/model/custom_vocab/base_manager/base_class_manager.py
```diff
@@ -33,66 +104,130 @@ def _get_new_custom_concept_class_ids(self) -> List[str]:
...
+
+                transformation_metadata = EtlTransformation(name=f'load_{class_file.stem}')
+
+                with self.db.session_scope(metadata=transformation_metadata) as session, \
+                        class_file.open('r') as f_in:
+                    rows = csv.DictReader(f_in, delimiter='\t')
+
+                    records = []
+
+                    for _, row in enumerate(rows, start=2):
```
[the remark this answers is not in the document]
alepev: replaced in all base managers

src/omop_etl_wrapper/model/custom_vocab/base_manager/base_class_manager.py
```diff
@@ -33,66 +104,130 @@ def _get_new_custom_concept_class_ids(self) -> List[str]:
...
+    def _update_custom_classes(self) -> None:
+        # Update the name of existing custom concept_classes in the
+        # database
+
+        classes_to_update = self._custom_classes_to_update
+
+        logging.info(f'Updating custom class names: '
+                     f'{True if classes_to_update else False}')
+
+        if classes_to_update:
```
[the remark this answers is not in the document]
alepev: suggestion taken

src/omop_etl_wrapper/model/custom_vocab/custom_vocab_loader.py
```diff
@@ -26,26 +26,57 @@ def __init__(self, db: Database, cdm):
...
-        # get custom vocab files for a specific vocabulary target table
-        # based on the file name conventions (e.g. "concept")
-        custom_table_files = self._get_all_custom_table_files()
+    def _get_custom_table_files(omop_table: str) -> List[Path]:
+        # Get custom vocab files for a specific vocabulary target table
+        # based on the file name conventions (e.g. "concept").
+        custom_table_files = get_all_files_in_dir(CUSTOM_VOCAB_DIR)
         return [f for f in custom_table_files if f.stem.endswith(omop_table)]
 
+    def _update_custom_files(self, file_list: List[Path], omop_table: str) -> List[Path]:
```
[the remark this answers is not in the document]
alepev: right, also updated

src/omop_etl_wrapper/model/custom_vocab/base_manager/base_vocab_manager.py
```diff
@@ -17,83 +21,192 @@ def __init__(self, db: Database, cdm, custom_vocab_files: List[Path]):
...
-                .filter(self._cdm.Vocabulary.vocabulary_id == vocab_id) \
-                .one_or_none()
-            return existing_record.vocabulary_version if existing_record is not None else None
+        for new_id, new_version in vocab_new.items():
+            old_version = vocab_old.get(new_id, None)
+
+            # skip version if already present in database
+            if old_version == new_version:
+                unchanged_vocabs.add(new_id)
+                continue
```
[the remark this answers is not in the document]
alepev: funny, no idea how that ended up there!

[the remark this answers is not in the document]
alepev: well, actually I do.. but it's not needed anymore

src/omop_etl_wrapper/model/custom_vocab/custom_vocab_loader.py
alepev: @Spayralbe I resolved this by creating a helper function in `io.py`: now the code is really the same for stcm and custom vocabulary managers https://github.com/thehyve/delphyne/blob/f30ab014b5d9c7d1d87d6cdf28aa69932f242022/src/omop_etl_wrapper/util/io.py#L38-L67

src/omop_etl_wrapper/model/custom_vocab/custom_vocab_loader.py
```diff
@@ -5,47 +5,77 @@
...
-        self._custom_concept_files = self._subset_custom_table_files('concept')
-        self._custom_class_files = self._subset_custom_table_files('concept_class')
+        self._custom_vocab_files: List[str] = []
+        self._custom_class_files: List[str] = []
+        self._custom_concept_files: List[str] = []
 
-        BaseVocabManager.__init__(self, db=self.db, cdm=self._cdm,
+    def _initialize_table_managers(self) -> None:
+        if not CUSTOM_VOCAB_DIR.exists():
+            raise FileNotFoundError(f'{CUSTOM_VOCAB_DIR.resolve()} folder not found')
```
alepev, line 23: @Spayralbe and also added this check, I realized it was missing while going over the stcm code again. 
I had to move the code in the `__init__` to a new method (`_initialize_table_managers`) to make sure the check is triggered only if the configuration option `skip_custom_vocabulary_loading` is off

alepev:
@Spayralbe I'm done updating and re-reviewing my own code :P aside from the two comments above, I was wondering what are the implications for the current updates and deletions code once we merge with `expand_listeners`.. as far as I can tell I won't need to refactor anything here, right?

src/omop_etl_wrapper/model/custom_vocab/custom_vocab_loader.py
Spayralbe: Yes, good addition!

Spayralbe:
> @Spayralbe I'm done updating and re-reviewing my own code :P aside from the two comments above, I was wondering what are the implications for the current updates and deletions code once we merge with `expand_listeners`.. as far as I can tell I won't need to refactor anything here, right?

Nice work! The new listeners should indeed be able to detect these transformations, so feel free to merge.

Spayralbe (approved):

PR #85, opened 2021-01-03
STCM tests

### LINKED ISSUE #61: Add tests for StcmLoader

Spayralbe:
- Closes #61 
- Refresh cached properties when loading STCM files
- Add `SourceToConceptMapVersion` table to test CDMs


tests/python/database/constraints/constraint_sets.py
```diff
@@ -158,6 +158,7 @@
     'fk_source_to_concept_map_source_vocabulary_id_vocabulary',
     'fk_source_to_concept_map_target_concept_id_concept',
     'fk_source_to_concept_map_target_vocabulary_id_vocabulary',
+    'fk_source_to_concept_map_version_source_vocabulary_id_v_f725',
```
alepev, line 161: I'm a bit confused by the name of this one

alepev (commented):
2 things:
- might be worth adding a test for mismatches between stcm file name and vocab ids used within the file. there are already suitable test files for this: `tests/test_data/stcm/stcm2`. I seem to remember the functionality that checks for this in the stcm code was already (partially?) implemented, see discussion in https://github.com/thehyve/delphyne/issues/57
- I was wondering about the mix of .tsv and .csv files. is .csv is something recommended in the OHDSI guidelines for stcm files, or perhaps to make them look like Usagi output files? just curious because I don't remember how we got there :P

tests/python/database/constraints/constraint_sets.py
Spayralbe: So was I 😅. But it's explained in the docs:

> When a generated name, particularly those that use the multiple-column tokens, is too long for the identifier length limit of the target database (for example, PostgreSQL has a limit of 63 characters), the name will be deterministically truncated using a 4-character suffix based on the md5 hash of the long name.

alepev: bizarre

PR #86, opened 2021-01-04
Expand event listeners

### LINKED ISSUE #82: Investigate how to capture deletions and updates
Currently only insertions are reported in `etl_stats`. How to get counts for updates and deletions when using session.query? Can we think of workarounds?

Spayralbe:
Depends on #81 

- Closes #82 
- Move all SQLAlchemy event listener methods from `database.py` into a separate `events.py` module.
- Replace Database.session property with a more intuitive `get_new_session` method.
- Change type of Database.schemas from set to frozenset to emphasize the contents are final.


wal004:
> Depends on #81

Would you be up for automatically managing dependent PRs? I built [Dpulls](https://github.com/marketplace/dpulls) to handle that - could be helpful if you often deal with dependent PRs - totally free for open source.

src/omop_etl_wrapper/database/events.py
```diff
@@ -0,0 +1,78 @@
...
+
+@event.listens_for(Session, 'after_bulk_update')
+def receive_after_bulk_update(update_context: BulkUpdate):
+    _process_bulk_event(update_context)
+
+
+@event.listens_for(Session, 'after_bulk_delete')
+def receive_after_bulk_delete(delete_context: BulkDelete):
+    print(f'Deleted {delete_context.rowcount} {delete_context.primary_table.fullname}')
+    _process_bulk_event(delete_context)
```
alepev, line 41: just a question about these: how/when do these methods get used in practice since I cannot find any reference to them elsewhere in the code? guess it's a magic trick of `@event.listerns_for`?

src/omop_etl_wrapper/database/events.py
```diff
@@ -0,0 +1,78 @@
...
+        target table
+    """
+    for record in record_containing_object:
+        placeholder_schema = record.__table__.schema
+        schema_name = Database.schema_translate_map.get(placeholder_schema, placeholder_schema)
+        table_name = record.__tablename__
+        if schema_name is None:
+            yield table_name
+        else:
+            yield '.'.join([schema_name, table_name])
```
alepev, line 61: could this also make use of  the `get_full_table_name` helper function?

alepev (approved):
Approved, though very curious to better understand how `@event.listens_for` works. Shall we also add tests for that (in a separate issue)?

src/omop_etl_wrapper/database/events.py
Spayralbe: Correct, the methods are not supposed to be called directly. They automatically trigger whenever the specific session event occurs.

src/omop_etl_wrapper/database/events.py
Spayralbe: Good point. Fixed now.

Spayralbe:
> Approved, though very curious to better understand how `@event.listens_for` works. Shall we also add tests for that (in a separate issue)?

Yes, a nice way to test them might be by checking whether the various listeners correctly capture events that are expected to be represented in `EtlTransformation` instances.

PR #88, opened 2021-01-07
Rename package to delphyne

### LINKED ISSUE #84: Naming the repository
After lengthy discussion and a poll, I think we have two come down to two good candidates, both from Greek mythology:
* Hellen/Hellenist, referring to the origin of the Greek mythology and for Hellenist a link to the Arborist
* Delphyne, referring to Python

Let's make a choice in the new year.

Spayralbe:
Closes #84 

Change all references from omop_etl_wrapper to delphyne.
- package and imports
- setup.py
- README


alepev (approved):
I couldn't find anything with the old name - except in `.gitignore`-d folders like `.nox` and similar. All good!

PR #91, opened 2021-01-11
Standard vocabulary loading

### LINKED ISSUE #15: Implement loading of standard OMOP vocabularies
Will require creating the vocabularies tables without constraints and indices.

Spayralbe:
Closes #15 depends on https://github.com/thehyve/delphyne-template/pull/55.

Add `StandardVocabLoader` to the `VocabManager` for loading the standard OMOP CDM vocabularies as downloaded from Athena (currently only for postgres, see #23).


src/delphyne/model/standard_vocab/standard_vocab_loader.py
```diff
@@ -0,0 +1,109 @@
...
+        finally:
+            connection.close()
+        transformation_metadata.end_now()
+        etl_stats.add_transformation(transformation_metadata)
+
+    def _check_vocab_tables_are_empty(self) -> None:
+        # We require all vocabulary tables to be empty beforehand, to
+        # avoid (time-consuming) accidental reloading.
+        for table_name in _STANDARD_VOCAB_TABLE_NAMES:
+            full_table_name = f'{VOCAB_SCHEMA}.{table_name}'
```
alepev, line 96: I'm a bit confused by this one. how will this match the actual vocabulary schema name at runtime, if using a placeholder name?

src/delphyne/model/standard_vocab/standard_vocab_loader.py
```diff
@@ -0,0 +1,109 @@
...
+        if not STANDARD_VOCAB_DIR.exists():
+            raise FileNotFoundError(f'standard vocabulary folder not found. '
+                                    f'Expected at {STANDARD_VOCAB_DIR}')
+        return set(get_all_files_in_dir(STANDARD_VOCAB_DIR))
+
+    def _create_table_file_mapping(self) -> None:
+        # Find the corresponding data file for each vocab table
+        for table_name in _STANDARD_VOCAB_TABLE_NAMES:
+            for vocab_file in self._standard_vocab_files:
+                if vocab_file.stem.lower() == table_name:
```
alepev, line 68: good you are checking for lowercase, I'm afraid we didn't do this in the custom vocab & stcm loaders 🤔

src/delphyne/model/standard_vocab/standard_vocab_loader.py
```diff
@@ -0,0 +1,109 @@
...
+        try:
+            cursor = connection.cursor()
+            statement = f"COPY {table} FROM STDIN WITH DELIMITER E'\t' CSV HEADER QUOTE E'\b';"
+            with vocab_file.open('rb') as f:
+                cursor.copy_expert(sql=statement, file=f)
+            transformation_metadata.insertion_counts += Counter({table: cursor.rowcount})
+            cursor.close()
+            connection.commit()
+        finally:
+            connection.close()
```
alepev, line 88: is there a suitable exception log if this goes wrong?

tests/python/model/standard_vocab/test_standard_vocab.py
```diff
@@ -0,0 +1,95 @@
...
+import tests.python.cdm.cdm600 as cdm
+from tests.python.conftest import docker_not_available
+
+pytestmark = pytest.mark.skipif(condition=docker_not_available(),
+                                reason='Docker daemon is not running')
+
+
+@pytest.fixture(scope='session')
+def base_standard_vocab_dir(test_data_dir: Path) -> Path:
+    return test_data_dir / 'standard_vocabs'
```
alepev, line 19: is this used anywhere?

alepev: nevermind

tests/python/model/standard_vocab/test_standard_vocab.py
```diff
@@ -0,0 +1,95 @@
...
+        with pytest.raises(FileNotFoundError, match=message):
+            wrapper.vocab_manager.load_standard_vocabularies()
+
+    # Temporarily remove the Base metadata, to check whether an
+    # exception is raised about vocabulary tables not being part of
+    # Base. Restore the original metadata afterwards.
+    original_metadata = copy(wrapper.db.base.metadata)
+    wrapper.db.base.metadata = MetaData()
+    with mock_standard_vocab_paths(base_standard_vocab_dir, 'vocab1'):
+        with pytest.raises(ValueError, match='Missing table "concept"'):
```
alepev, line 62: if I understand correctly, all metadata are removed, but the error is specifically for `concept` because that's the first table in alphabetical order.. right? I guess we can trust this to always be the case

tests/python/model/standard_vocab/test_standard_vocab.py
```diff
@@ -0,0 +1,95 @@
...
+    return test_data_dir / 'standard_vocabs'
+
+
+@contextmanager
+def mock_standard_vocab_paths(base_dir: Path, standard_vocab_dir_name: str):
+    """Mock global variable STANDARD_VOCAB_DIR."""
+    vocab_dir = base_dir / standard_vocab_dir_name
+    import_path = 'src.delphyne.model.standard_vocab.standard_vocab_loader.STANDARD_VOCAB_DIR'
+    with patch(import_path, vocab_dir):
+        yield
```
alepev, line 28: I was wondering why we need to mock the path like this

alepev:
Aaaaah how did this guy get committed?!?
https://github.com/thehyve/delphyne/blob/55f09d97accf6fedd379a80c86ba413b4472e647/tests/test_data/standard_vocabs/.DS_Store

src/delphyne/model/standard_vocab/standard_vocab_loader.py
alepev: I added a separate issue for that: #95

alepev (approved):
Minor comments, including getting rid of `.DS_Store` :P

src/delphyne/model/standard_vocab/standard_vocab_loader.py
Spayralbe: It doesn't. In the metadata.tables dict-like object, the keys are the full table names, of which the schema names are the placeholder schema names. Only when using these tables in a database interaction will they be replaced with the actual schema names as bound to the engine.

src/delphyne/model/standard_vocab/standard_vocab_loader.py
Spayralbe: You will get the original postgres error message encapsulated in a psycopg2 exception.

tests/python/model/standard_vocab/test_standard_vocab.py
Spayralbe: It's indeed not the optimal way to test for this exception, but creating a metadata instance with specifically one vocabulary table missing seemed more trouble than it's worth :p

tests/python/model/standard_vocab/test_standard_vocab.py
Spayralbe: In a regular run, `STANDARD_VOCAB_DIR` is a constant and the only path at which you can provide standard vocabulary files. For testing we need to mock it, because otherwise the path doesn't exist and we would have no way to test it with different input files.

Spayralbe:
> Aaaaah how did this guy get committed?!?
> https://github.com/thehyve/delphyne/blob/55f09d97accf6fedd379a80c86ba413b4472e647/tests/test_data/standard_vocabs/.DS_Store

Argh sorry, bloody DS_Store! 🙈. I'll add it to the gitignore straight away :p

src/delphyne/model/standard_vocab/standard_vocab_loader.py
alepev: OK, I think it's clear now

tests/python/model/standard_vocab/test_standard_vocab.py
alepev: I will have to remember this one, probably will be needed in other tests

PR #93, opened 2021-01-11
EtlStats reporter

### LINKED ISSUE #90: Failed transformations summary log not printed
The `_log_transformations` method only prints out successful transformations, even if some transformations failed.

My suspicion is that the failed transformation stats are filtered out here (because it `is_empty`).
https://github.com/thehyve/delphyne/blob/b9d5d300d60d85406701b8f09d29711eb5c56e78/src/delphyne/model/etl_stats/etl_stats_reporter.py#L48-L49

Spayralbe:
- Fixes #90 
- Add basic tests on what `EtlTransformation` instances are expected to end up in the summary report.

### DIFF
--- a/tests/python/model/etl_stats/test_etl_stats_reporter.py
+++ b/tests/python/model/etl_stats/test_etl_stats_reporter.py
@@ -1,0 +1,53 @@
+import logging
+from collections import Counter
+
+import pytest
+from src.delphyne.model.etl_stats import EtlStats, EtlTransformation, EtlStatsReporter
+
+from tests.python.model.etl_stats.conftest import get_etltransformation
+
+logger = logging.getLogger('test_logger')
+logger.setLevel(logging.DEBUG)
+
+
+@pytest.fixture(scope='function')
+def reporter_summary_output(request, caplog) -> str:
+    """Get summary report for the given EtlTransformation."""
+    stats = EtlStats()
+    transformation: EtlTransformation = request.param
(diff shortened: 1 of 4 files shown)


tests/python/model/etl_stats/test_etl_stats_reporter.py
```diff
@@ -0,0 +1,53 @@
...
+    stats.add_transformation(transformation)
+    reporter = EtlStatsReporter(etl_stats=stats)
+    with caplog.at_level(logging.INFO):
+        reporter.log_summary()
+    return caplog.text
+
+
+@pytest.mark.parametrize('reporter_summary_output',
+                         [get_etltransformation(name='empty_successful')],
+                         indirect=True)
```
alepev, line 27: curious: what is `indirect` for?

alepev (approved):

tests/python/model/etl_stats/test_etl_stats_reporter.py
Spayralbe: In this case, it's a way to specify you don't want to use the return value of `get_etltransformation` directly in your test, but first pass it into `reporter_summary_output` and get the output of that fixture. See [here](https://docs.pytest.org/en/stable/example/parametrize.html#indirect-parametrization).

alepev: thanks!

PR #96, opened 2021-01-12
Case insensitive file check

### LINKED ISSUE #95: Add lowercase check to stcm & custom vocab filenames
When checking file names, the suffix is currently compared as is. Convert it to lowercase before comparing.

alepev:
Closes #95

Makes comparison of custom vocabulary file names case-insensitive. 
Both prefixes and suffixes will be checked against vocabulary ids and table names (respectively) in lowercase.

Comparisons between vocabulary ids already loaded to the database and vocabulary ids provided in the files remain case-sensitive.

..Or do you think making the prefix check case-insensitive is a bad idea? 🤔 Debatable

### DIFF
--- a/src/delphyne/model/custom_vocab/base_manager/base_concept_manager.py
+++ b/src/delphyne/model/custom_vocab/base_manager/base_concept_manager.py
@@ -27,17 +27,18 @@ class BaseConceptManager:
         logging.info(f'Dropping old custom concepts: '
                      f'{True if vocab_ids else False}')
 
-        if vocab_ids:
+        if not vocab_ids:
+            return
 
-            transformation_metadata = EtlTransformation(name='drop_concepts')
+        transformation_metadata = EtlTransformation(name='drop_concepts')
 
-            with self._db.session_scope(metadata=transformation_metadata) as session:
-                session.query(self._cdm.Concept) \
-                    .filter(self._cdm.Concept.vocabulary_id.in_(vocab_ids)) \
-                    .delete(synchronize_session=False)
+        with self._db.session_scope(metadata=transformation_metadata) as session:
+            session.query(self._cdm.Concept) \
+                .filter(self._cdm.Concept.vocabulary_id.in_(vocab_ids)) \
+                .delete(synchronize_session=False)
 
-                transformation_metadata.end_now()
-                etl_stats.add_transformation(transformation_metadata)
+            transformation_metadata.end_now()
+            etl_stats.add_transformation(transformation_metadata)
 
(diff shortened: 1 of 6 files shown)


src/delphyne/model/raw_sql_wrapper.py
```diff
@@ -39,8 +39,9 @@ def execute_sql_file(self, file_path: Union[Path, str]) -> None:
         """
         Executes raw SQL file.
 
-        :param file_path: relative SQL file path inside the directory for
-            SQL transformations (the root will be automatically added).
+        :param file_path: relative SQL file path inside the directory
+            for SQL transformations
+            (the root will be automatically added).
```
alepev, line 44: this is an unrelated change because of the linter complaining.. even though the text was just within the 72 character line, according to my eyes 🤷

src/delphyne/util/io.py
```diff
@@ -28,7 +28,20 @@ def get_all_files_in_dir(directory: Path) -> List[Path]:
 
 
 def get_file_prefix(path: Path, suffix: str) -> Optional[str]:
-    stem = path.stem
+    """
+    Extracts the part of the file name preceding the provided suffix.
```
Spayralbe, line 32: From pep257:

> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

Spayralbe:
```suggestion
    Extract the part of the file name preceding the provided suffix.
```

src/delphyne/model/custom_vocab/base_manager/base_concept_manager.py
```diff
@@ -90,9 +90,11 @@ def _load_custom_concepts(self, vocab_ids: Set[str], valid_prefixes: Set[str]) -
...
                             ))
 
-                        # if prefix is valid vocab_id,
-                        # vocabulary_ids in file should match it
-                        if prefix in valid_prefixes and vocabulary_id != prefix:
+                        # if file prefix is valid vocab_id,
+                        # vocabulary_ids in file should match it.
+                        # comparison is case-insensitive.
+                        vocabs_lowercase = {vocab.lower() for vocab in valid_prefixes}
+                        if file_prefix in vocabs_lowercase and vocabulary_id.lower() != file_prefix:
```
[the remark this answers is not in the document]
alepev: reduced indentation in these an other custom vocab files too.. not sure what's going on with my IDE, but I reduced the length of visual guides by 1 char. that should help

Spayralbe (approved):

PR #103, opened 2021-01-13
Allow vocabs without concepts

### LINKED ISSUE #100: Allow vocabulary-only custom vocabularies
A call to `load_custom_vocabularies` currently doesn't allow adding or updating vocabularies that consist of only a record in the vocabulary table (no linked concept_ids).

https://github.com/thehyve/delphyne/blob/37afdf7496577214749eccccdaa68c6eeb2d3c50/src/delphyne/model/custom_vocab/custom_vocab_loader.py#L68-L70

For STCM files it is however rather common to only need a custom source_vocabulary_id value, but no linked source_concept_ids. It would therefore be useful to allow this scenario and load only the vocabulary_id records.

alepev:
Closes #100

### DIFF
--- a/src/delphyne/model/custom_vocab/custom_vocab_loader.py
+++ b/src/delphyne/model/custom_vocab/custom_vocab_loader.py
@@ -64,11 +64,6 @@ class CustomVocabLoader(BaseVocabManager, BaseClassManager, BaseConceptManager):
         # since not particularly large)
         self._custom_concept_files = self._update_custom_file_list(
             self._custom_concept_files, 'concept')
-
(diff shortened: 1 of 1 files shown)


alepev:
I was just wondering, how did the STCM tests work despite this? 😅

Spayralbe:
> I was just wondering, how did the STCM tests work despite this? 😅

To test them (somewhat) independently, they don't make use of the `CustomVocabLoader`. But it would definitely be useful to have an additional test that combines these components (as this issue shows :p).

Spayralbe (approved):

PR #104, opened 2021-01-13
Use contextmanager for EtlTransformations

Spayralbe:
- Simplifies tracking changes via `EtlTransformation` by using a contextmanager. That way there is no need to manually set start or end times and instances are automatically added to the `etl_stats` collection.
- Add a `tracked_session_scope` method to `Database` using the previous as part of a session. That way both the session and the `EtlTransformation` instances can be managed in a single with statement.

### DIFF
--- a/src/delphyne/database/database.py
+++ b/src/delphyne/database/database.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 from contextlib import contextmanager
 from getpass import getpass
-from typing import Dict, Optional, Set, FrozenSet
+from typing import Dict, Set, FrozenSet, ContextManager, Tuple
 
 from sqlalchemy import create_engine, MetaData
 from sqlalchemy.exc import OperationalError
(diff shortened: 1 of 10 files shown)


src/delphyne/model/custom_vocab/base_manager/base_vocab_manager.py
```diff
@@ -143,16 +142,11 @@ def _drop_custom_vocabs(self) -> None:
         if not vocabs_to_drop:
             return
 
-        transformation_metadata = EtlTransformation(name='drop_concepts')
-
-        with self._db.session_scope(metadata=transformation_metadata) as session:
+        with self._db.tracked_session_scope(name='drop_concepts') as (session, _):
```
alepev, line 145: I made a mistake here: the transformation name should be `drop_vocabs`

src/delphyne/database/database.py
```diff
@@ -90,38 +90,82 @@ def perform_rollback(session: Session) -> None:
...
+            SQLAlchemy open session.
+        EtlTransformation
+            Container storing metadata about the session.
+        """
+        session = self.get_new_session()
+        session_id = id(session)
+        with open_transformation(name=name) as metadata:
+            SessionTracker.sessions[session_id] = metadata
+            try:
+                yield session, metadata
```
alepev, line 157: stupid question: is `yield` like a `return` statement, except it allows you to perform other operations before you actually return the specified variables?

src/delphyne/model/custom_vocab/base_manager/base_vocab_manager.py
Spayralbe: Fixed

PR #113, opened 2021-01-18
Docstrings

### LINKED ISSUE #99: Standardize docstrings
Make sure all docstrings adhere to the [NumPy style](https://numpydoc.readthedocs.io/en/latest/format.html).

Spayralbe:
- Add NumPy docstrings to all public modules, methods and classes (Closes #99)
- Make `schema_translate_map` immutable
- Restructure `VocabManager` to call vocab loading from composition classes directly
- Add docstring linter to nox

### DIFF
--- a/src/delphyne/database/database.py
+++ b/src/delphyne/database/database.py
@@ -1,3 +1,5 @@
+"""Database operations module."""
+
 from __future__ import annotations
 
 import logging
@@ -42,6 +44,7 @@ class Database:
     constraint_manager : ConstraintManager
         Access point to alter constraints/indexes of the database.
     """
+
     schema_translate_map: MappingProxyType = None
 
     def __init__(self, uri: str, schema_translate_map: Dict[str, str], base):
@@ -212,8 +215,7 @@ class Database:
     @staticmethod
     def can_connect(uri: str) -> bool:
         """
-        Check whether a database connection can be established for the
-        given URI.
+        Check whether a connection can be established for the given URI.
 
         Parameters
         ----------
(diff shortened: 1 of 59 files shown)


src/delphyne/database/database.py
```diff
@@ -1,8 +1,11 @@
+"""Database operations module."""
```
alepev, line 1: Do we need to add strings like this at the top of every file (including `__init__`s) from now on?

src/delphyne/database/database.py
```diff
@@ -20,10 +23,32 @@
...
+
+    Attributes
+    ----------
+    schemas
+    reflected_metadata
+    engine : sqlalchemy.engine.base.Engine
+        Database engine.
+    constraint_manager : ConstraintManager
+        Access point to alter constraints/indexes of the database.
+    """
```
alepev, line 46: I think I've seen `Parameters` described for all classes; not sure if it's always the case for `Attributes`, and if/when it's really needed. Do you follow specific guidelines for this? Just curious to know if I should document them in the future

alepev (approved):

src/delphyne/database/database.py
Spayralbe: Yes, or the linter police will know where to find you.

src/delphyne/database/database.py
Spayralbe: I understand it as any public attribute that is not part of the `init`(those are already described under `Parameters`)
[https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring](https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring )

PR #130, opened 2021-01-21
Add combined method for custom vocabs and stcm

### LINKED ISSUE #128: Create combined STCM and custom vocab method in VocabManager
Currently, when a new version of a custom vocabulary is provided, the records of that vocabulary are first deleted and then the new version is loaded. This breaks FK constraints when the vocabulary was already used by an STCM mapping. To make it work, you therefore need to drop the STCM records manually first.

~~It would be more convenient if these STCM FKs were dropped in case vocabularies need to be reloaded that are used by the STCM tables. They should then be reapplied afterwards.~~

Spayralbe:
Closes #128

### DIFF
--- a/src/delphyne/model/vocab_manager.py
+++ b/src/delphyne/model/vocab_manager.py
@@ -38,7 +38,7 @@ class VocabManager:
         self.custom_vocabularies = CustomVocabLoader(db, cdm, skip_custom_vocabs)
         self.stcm = StcmLoader(db, cdm, skip_stcm)
 
-    def process_updates(self) -> None:
+    def load_custom_vocab_and_stcm_tables(self) -> None:
         """
         Process any updates in custom vocabulary/STCM files.
 
--- a/src/delphyne/model/vocab_manager.py
+++ b/src/delphyne/model/vocab_manager.py
@@ -52,14 +52,19 @@ class VocabManager:
         """
         stcm_table_name = self._cdm.SourceToConceptMap.__tablename__
         stcm_version_table_name = self._cdm.SourceToConceptMapVersion.__tablename__
-
-        self._db.constraint_manager.drop_table_constraints(stcm_table_name,
-                                                           drop_pk=False, drop_index=False)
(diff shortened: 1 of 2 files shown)


Spayralbe:
@alepev If you have a better idea for naming the method, I'm open to suggestions :p

src/delphyne/model/vocab_manager.py
```diff
@@ -27,10 +27,39 @@ class VocabManager:
...
+
         skip_standard_vocabs = config.run_options.skip_vocabulary_loading
         skip_custom_vocabs = config.run_options.skip_custom_vocabulary_loading
         skip_stcm = config.run_options.skip_source_to_concept_map_loading
 
         self.standard_vocabularies = StandardVocabLoader(db, cdm, skip_standard_vocabs)
         self.custom_vocabularies = CustomVocabLoader(db, cdm, skip_custom_vocabs)
         self.stcm = StcmLoader(db, cdm, skip_stcm)
+
+    def process_updates(self) -> None:
```
alepev, line 41: not ugly but maybe a bit vague. is it too long if we call it something explicit like `load_custom_vocab_and_stcm_tables`?

alepev: but then for name consistency we might also add back another method to call the standard vocabularies, instead of `self.vocab_manager.standard_vocabularies.load()` 🤔

alepev (commented):
Code is good, for the name we can think a bit more :P

src/delphyne/model/vocab_manager.py
Spayralbe: I agree with the method name change. Shows the behavior more clearly.

The latter point I'm bit more hesitant about. It would mean we need to add a separate method to VocabManager for each public method in any of the Loader classes used in composition (and duplicate the docstrings). I'd argue having a subdivision per Loader class would make things easier to find (especially if we would add more public methods later) than if they're all directly available in the VocabManager namespace.

Spayralbe: We decided to only update the name for now

PR #133, opened 2021-01-25
Constraint errors option

### LINKED ISSUE #131: Gracefully handle errors when applying constraints
# Current behaviour
The `add_cdm_constraints` will exit on the first constraint that can't be applied. Part of the constraints will be applied (the ones before the error), but the others will not be applied. Also we don't know if there are other constraints that would fail.

# Expected behaviour
Gracefully handle errors for individual applied constraints. Only display a warning for the constraints that could not be applied and continue with the other constraints.

# Proposed solution
The constraints are already applied individually. So we can easily solve this by adding a `try: except:` around `_add_constraint_in_db` and log the error message as warning or error:
https://github.com/thehyve/delphyne/blob/7f4879e939b033845207e498d72c05d8bfbfda4e/src/delphyne/database/constraints/constraint_manager.py#L296-L298

Spayralbe:
- Resolves #131 by adding an `errors`option to the constraint calls, which allow you to specify what should happen in case a constraint/index cannot be added/dropped.
- No longer separate methods for adding/dropping a single index or constraint. Only `drop_constraint` and `add_constraint`.

### DIFF
--- a/tests/python/database/constraints/test_constraints.py
+++ b/tests/python/database/constraints/test_constraints.py
@@ -3,6 +3,7 @@ from typing import Set
 
 import pytest
 from sqlalchemy import Table, Index, Constraint, MetaData
+from sqlalchemy.exc import InternalError, ProgrammingError
 from src.delphyne import Wrapper
 
 from tests.python.conftest import docker_not_available
@@ -54,14 +55,14 @@ def test_drop_and_add_single_index(cdm531_wrapper_with_tables_created: Wrapper):
     expected_full = expected_sets.measurement_indexes_full
     assert get_index_names(meas_table.indexes) == expected_full
 
-    # Calling drop_index will remove the index on person_id
-    wrapper.db.constraint_manager.drop_index(person_id_index)
+    # Calling drop_constraint will remove the index on person_id
+    wrapper.db.constraint_manager.drop_constraint(person_id_index)
     meas_table = reflect_table(wrapper, full_table_name)
     expected = expected_sets.measurement_indexes_without_person_index
     assert get_index_names(meas_table.indexes) == expected
 
     # Calling add_index will restore the index on person_id
-    wrapper.db.constraint_manager.add_index(person_id_index)
+    wrapper.db.constraint_manager.add_constraint(person_id_index)
     meas_table = reflect_table(wrapper, full_table_name)
     assert get_index_names(meas_table.indexes) == expected_full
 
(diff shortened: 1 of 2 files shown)


tests/python/database/constraints/test_constraints.py
```diff
@@ -97,16 +98,31 @@ def test_raise_keyerror_if_object_not_found(cdm600_wrapper_with_tables_created:
...
+    # If a valid constraint cannot be added/dropped, raise an exception,
+    # unless specified otherwise.
+    with pytest.raises(InternalError):
+        wrapper.db.constraint_manager.drop_constraint('pk_person')
+    wrapper.db.constraint_manager.drop_constraint('pk_person', errors='ignore')
+
+    wrapper.db.constraint_manager.drop_cdm_constraints()
+    with pytest.raises(ProgrammingError):
+        wrapper.db.constraint_manager.add_table_constraints('observation')
+    wrapper.db.constraint_manager.add_table_constraints('observation', errors='ignore')
```
alepev, line 125: in these blocks, I understand the part inside the `with` block means `pytest` is checking that the raised error is the expected one; but I don't get how you can test the part after the `with` block (where you repeat the call by passing `errors='ignore'`), since there is no `assert` or other pytest-specific function 🤔

alepev (commented):
Nice work! I have a question in the tests, and a general suggestion: it's good to use methods for both constraints and indexes to avoid code duplication, but shall we call them `add_constraint_or_index` / `drop_constraint_or_index` (for example) instead of only using the word `constraint` in the name? the docstrings make very clear that you can use these methods for indexes, but this can make it even clearer.. or am I being too picky? 😅

tests/python/database/constraints/test_constraints.py
Spayralbe: The only thing being tested there is that the call does not raise any exceptions. If it would, the test would fail.

Spayralbe:
> Nice work! I have a question in the tests, and a general suggestion: it's good to use methods for both constraints and indexes to avoid code duplication, but shall we call them `add_constraint_or_index` / `drop_constraint_or_index` (for example) instead of only using the word `constraint` in the name? the docstrings make very clear that you can use these methods for indexes, but this can make it even clearer.. or am I being too picky? 😅

No you're right. If only there was a word that encompasses both constraints and indexes, but not other things 🤔

tests/python/database/constraints/test_constraints.py
alepev: ok, I was overthinking it

PR #138, opened 2021-02-04
Batch insert

### LINKED ISSUE #137: Memory optimisation with batch inserts
**How would this feature be useful?**
When inserting many, many records from one transformation all the records are kept in a list before inserting. This will use a lot of memory. Ideally we intermittently insert the records to the database.

**Describe the solution you'd like**
An idea is to define the transformation function as a generator:
```python
def source_to_target():
    for s in df:
        # transformations
        yield Observation(...)
```

Then in delphyne:
```python
def execute_transformation(self, statement, bulk, batch_size = 10_000):
    records_to_insert = []
    for record in statement():
        records_to_insert.append(record)
        if len(records) >= batch_size:
            logger.info(f'Executing transformation: {batch_size} {statement.__name__} ')
            
            logger.info(f'Saving {len(records_to_insert)} objects')
            if bulk:
                session.bulk_save_objects(records_to_insert)
                self._collect_query_statistics_bulk_mode(session, records_to_insert,
                                                         transformation_metadata)
            else:
                session.add_all(records_to_insert)

        logger.info(f'{statement.__name__} completed with success status: '
                        f'{transformation_metadata.query_success}')
```

MaximMoinat:
Fixes #137

### DIFF
--- a/src/delphyne/model/orm_wrapper.py
+++ b/src/delphyne/model/orm_wrapper.py
@@ -54,6 +54,8 @@ class OrmWrapper(ABC):
     def execute_batch_transformation(self, batch_statement: Callable, bulk: bool = False, batch_size: int = 10000) -> None:
         """
         Execute an ETL transformation in batches of given size via a python statement.
+        Batches are committed to the database independently and a failed insertion of a batch will
+        not trigger a rollback of the other batches.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


src/delphyne/model/orm_wrapper.py
```diff
@@ -51,15 +51,81 @@ def is_git_repo() -> bool:
...
+                    logger.info(f'Saving {len(records_to_insert)} objects (batch {batch_count}): '
+                                f'{batch_statement.__name__} ')
+                    if bulk:
+                        session.bulk_save_objects(records_to_insert)
+                        self._collect_query_statistics_bulk_mode(session, records_to_insert,
+                                                                 transformation_metadata)
+                    else:
+                        session.add_all(records_to_insert)
+                    # Both flushing and committing to write from memory to disk
+                    session.commit()  # also flushes: https://docs.sqlalchemy.org/en/14/orm/session_basics.html#committing
```
MaximMoinat, line 100: We could also consider setting up a new session for each batch insert. Not sure how that would affect performance and retrieving insert statistics.

src/delphyne/model/orm_wrapper.py
```diff
@@ -51,15 +51,88 @@ def is_git_repo() -> bool:
...
+
+        Returns
+        -------
+        None
+        """
+        logger.info(f'Executing batched transformation: {batch_statement.__name__} ')
+        func_args = signature(batch_statement).parameters
+        if 'session' in func_args:
+            session = self.db.get_new_session()
+            records_generator = batch_statement(self, session)
```
MaximMoinat, line 80: @Spayralbe Every batch now uses its own session. So also this session passed to the statement should be separate. Not sure what consequences this could have.

src/delphyne/model/orm_wrapper.py
```diff
@@ -51,15 +51,93 @@ def is_git_repo() -> bool:
         """
         return os.path.exists('./.git')
 
+    def execute_batch_transformation(self, batch_statement: Callable, bulk: bool = False, batch_size: int = 10000) -> None:
+        """
+        Execute an ETL transformation in batches of given size via a
+        python statement. Batches are committed to the database
+        independently and a failed insertion of a batch will
+        not trigger a rollback of the other batches.
+
```
MaximMoinat, line 60: @Spayralbe For me, this is the behaviour we want. It can give useful information while debugging (does the issue happen across the data, or just certain records?). On the other hand, this can result in an end state with just a partial transformation of the source table.

src/delphyne/model/orm_wrapper.py
Spayralbe: Yeah I would suggest to simplify this block:

```python
        func_args = signature(batch_statement).parameters
        if 'session' in func_args:
            session = self.db.get_new_session()
            records_generator = batch_statement(self, session)
        else:
            records_generator = batch_statement(self)
```

into only:
```python
records_generator = batch_statement(self)
```
Inserting a session like in `execute_transformation` will not have the same effect here, as objects added added to it will not be committed, so it's better to not allow it. If you really do need a separate session within the transformation, you can just create one yourself via the wrapper.

src/delphyne/model/orm_wrapper.py
Spayralbe: Agreed, best to keep it like this.

Spayralbe (commented):
A useful addition! I've added some suggestions in the comments.

Spayralbe (approved):

PR #139, opened 2021-02-04
Docs

### LINKED ISSUE #27: Create basic README.md
- [x] description
- [x] basic installation
- [x] dependencies
- [x] license
- [x] link to documentation (depends on #28 )
- [x] test instructions
- [x] ...

### LINKED ISSUE #28: Set up documentation
- [x] decide on Read the Docs, Just the Docs, other..
- [x] automatically generate basic documentation (public methods)

### LINKED ISSUE #118: Document loading of standard vocabulary tables
Until we have a method in the wrapper to do this, it's useful to provide directions to the user on how to load the standard Athena vocabularies after creating the tables. Relevant points are:
- column order is different in the ORM than it is in the Athena vocabularies, so column names should be made explicit when loading using a standard COPY from file SQL script. See here for an example: https://github.com/thehyve/ohdsi-etl-template/blob/custom_vocabularies/other/manual_vocabulary_tables_loading/vocab_loading_explicit.sql
- By default constraints and indexes are applied automatically during vocabulary table creation. This leads to error when executing the previous COPY statements because there are reciprocal dependencies between tables. We should either omit all constraints and explain how to apply them once manual loading is complete, or explain how to remove and reapply existing constraints as needed.

### LINKED ISSUE #119: Document custom vocabularies format and loading
See thehyve/delphyne-template#9 for general file conventions.

Additional notes:
- convention: `vocabulary_concept_id` and `concept_class_concept_id` should be always set to `0` for custom vocabularies/classes
- ~warning: updating a custom `vocabulary_id` will cause a loading error! the pipeline attempts to load the vocabulary as if it was completely new, without dropping the `concept_id`s previously associated with it. The loading fails because it would cause a duplicate PK in the `concept` table. **Long story short: if you update a custom vocabulary id, you need to manually drop existing `concept_id`s for that vocabulary before attempting a reload!**~ SOLVED
- ~warning: updating `concept_class_name` without updating the vocabulary version will cause a loading error! (the code tries to unload the old class version, but since the vocabularies didn't change, the custom concepts are not dropped first, and they will still reference the class when trying to drop it). **Long story short: if you update a custom class name, also update the vocabulary version!**~ SOLVED
- warning: updating `vocabulary_name` or `vocabulary_reference` without updating `vocabulary_version` will have no effect.
- ~warning: old `vocabulary_id`s and `concept_class_id`s won't be cleaned up automatically (until this issue is taken care of: https://github.com/thehyve/omop-etl-wrapper/issues/53)~ SOLVED

Spayralbe:
Closes #27 closes #28 closes #118 closes #119


docs/index.rst
```diff
@@ -0,0 +1,36 @@
+delphyne
```
Spayralbe, line 1: This file is a stub. More contents will need to be added here.

README.md
```diff
@@ -12,26 +15,31 @@
...
+delphyne is a Python package designed to simplify and standardize the process of converting
+source data into the [OMOP Common Data Model](https://www.ohdsi.org/data-standardization/the-common-data-model/).
 
+Full documentation is available on [Read the Docs](https://delphyne.readthedocs.io/en/latest/).
+
+## Main Features
+-   Customizable SQLAlchemy ORM of multiple OMOP CDM releases and extensions
+-   Flexible data transformation options using either raw SQL or the ORM
+-   Built-in tracking of ETL transformations to generate an overview of table record changes
+-   Comprehensive integration of source data files into transformations, with options for caching and various data formats
```
alepev, line 28: "integration" sounds a bit strange for source data, maybe "import"? really a minor thing

README.md
```diff
@@ -12,26 +15,31 @@
...
 
 ```sh
-git clone https://github.com/thehyve/delphyne.git
-cd delphyne
-
-# regular installation
-pip install .
-
-# or install in editable mode (including test dependencies)
-pip install -e '.[TEST]'
```
alepev, line 32: Is there another place to capture this (for internal use)? otherwise I just make a note somewhere

alepev: nevermind, I just saw it below

docs/reference.rst
```diff
@@ -0,0 +1,78 @@
...
+    :backlinks: none
+
+
+Wrapper
+-------
+
+.. autoclass:: src.delphyne.wrapper.Wrapper
+   :members:
+   :inherited-members:
+   :exclude-members: cdm
```
alepev, line 15: Nice, so this is how you can automatically generate the docs for a given class!
Does it automatically exclude private methods/properties?

docs/custom_vocab.rst
```diff
@@ -0,0 +1,63 @@
...
+::
+
+    project_root
+    └── resources
+        └── vocabularies
+            └── custom
+                ├── MYVOCAB_vocabulary.tsv
+                ├── MYVOCAB_concept.tsv
+                ├── vocabulary.tsv
+                └── etc.
```
alepev, line 35: I was wondering if it can be confusing to mix vocab files with/without the prefix, though it does show they are not mutually exclusive.. mmm

alepev (approved):
The docs so far read nicely :) I wonder if it's possible to add sub-sections if the table of contents becomes too long, and also if there's any way to sort the topics in custom rather than alphabetical order. but not for now, enough for the first iteration!

README.md
Spayralbe: It's a bit odd indeed :p I've rephrased it.

docs/reference.rst
Spayralbe: Yes those are excluded by default :)

Spayralbe:
> The docs so far read nicely :) I wonder if it's possible to add sub-sections if the table of contents becomes too long, and also if there's any way to sort the topics in custom rather than alphabetical order. but not for now, enough for the first iteration!

Thanks. Yes it should be possible to change those things to our liking.

PR #140, opened 2021-02-08
CDM v5.3.1 fixes

### LINKED ISSUE #112: Check `v5.3.1_fixes` branch of official CDM repo
Compare our current CDM v5.3.1 implementation to the `v5.3.1_fixes` branch in the official CDM repo and apply any changes: https://github.com/OHDSI/CommonDataModel/tree/v5.3.1_fixes/PostgreSQL

### LINKED ISSUE #132: Missing foreign key to visit_detail_id for v5.3.1
The official v5.3.1 release does not contain any foreign keys from the event tables to the `visit_detail` table. The [v5.3.1_fixes branch](https://github.com/OHDSI/CommonDataModel/tree/v5.3.1_fixes/PostgreSQL) DOES have these foreign keys.

I propose to follow the cdm definition in the fixes branch for v5.3.1. Also related to the cost typo fixes we did for #108.

e.g. in condition_occurrence table:
https://github.com/thehyve/delphyne/blob/7f4879e939b033845207e498d72c05d8bfbfda4e/src/delphyne/cdm/cdm531/clinical_data.py#L59-L61

alepev:
Fixes #112
Fixes #132

Depends on https://github.com/thehyve/delphyne-template/pull/64

### DIFF
--- a/src/delphyne/cdm/cdm531/clinical_data.py
+++ b/src/delphyne/cdm/cdm531/clinical_data.py
@@ -404,6 +404,24 @@ class BaseFactRelationshipCdm531:
     def relationship_concept_id(cls):
         return Column(ForeignKey(f'{VOCAB_SCHEMA}.concept.concept_id'), primary_key=True, nullable=False, index=True)
 
+    @declared_attr
+    def domain_concept_1(cls):
+        return relationship('Concept',
+                            primaryjoin='FactRelationship.domain_concept_id_1 == '
+                                        'Concept.concept_id')
+
+    @declared_attr
+    def domain_concept_2(cls):
+        return relationship('Concept',
+                            primaryjoin='FactRelationship.domain_concept_id_2 == '
+                                        'Concept.concept_id')
+
+    @declared_attr
+    def relationship_concept(cls):
+        return relationship('Concept',
+                            primaryjoin='FactRelationship.relationship_concept_id == '
+                                        'Concept.concept_id')
+
 
 class BaseMeasurementCdm531:
     __tablename__ = 'measurement'
--- a/src/delphyne/cdm/cdm531/clinical_data.py
+++ b/src/delphyne/cdm/cdm531/clinical_data.py
@@ -62,7 +62,7 @@ class BaseConditionOccurrenceCdm531:
 
     @declared_attr
     def visit_detail_id(cls):
-        return Column(Integer)
+        return Column(ForeignKey(f'{CDM_SCHEMA}.visit_detail.visit_detail_id'))
 
     @declared_attr
     def condition_source_value(cls):
@@ -103,6 +103,10 @@ class BaseConditionOccurrenceCdm531:
     @declared_attr
     def visit_occurrence(cls):
         return relationship('VisitOccurrence')
+
+    @declared_attr
+    def visit_detail(cls):
+        return relationship('VisitDetail')
 
 
 class BaseDeathCdm531:
@@ -208,7 +212,7 @@ class BaseDeviceExposureCdm531:
 
     @declared_attr
     def visit_detail_id(cls):
-        return Column(Integer)
+        return Column(ForeignKey(f'{CDM_SCHEMA}.visit_detail.visit_detail_id'))
 
     @declared_attr
     def device_source_value(cls):
@@ -241,6 +245,10 @@ class BaseDeviceExposureCdm531:
     @declared_attr
     def visit_occurrence(cls):
         return relationship('VisitOccurrence')
+
+    @declared_attr
+    def visit_detail(cls):
+        return relationship('VisitDetail')
 
 
 class BaseDrugExposureCdm531:
@@ -321,7 +329,7 @@ class BaseDrugExposureCdm531:
 
     @declared_attr
     def visit_detail_id(cls):
-        return Column(Integer)
+        return Column(ForeignKey(f'{CDM_SCHEMA}.visit_detail.visit_detail_id'))
 
     @declared_attr
     def drug_source_value(cls):
@@ -366,6 +374,10 @@ class BaseDrugExposureCdm531:
     @declared_attr
     def visit_occurrence(cls):
         return relationship('VisitOccurrence')
+
+    @declared_attr
+    def visit_detail(cls):
+        return relationship('VisitDetail')
 
 
 class BaseFactRelationshipCdm531:
@@ -459,7 +471,7 @@ class BaseMeasurementCdm531:
 
     @declared_attr
     def visit_detail_id(cls):
-        return Column(Integer)
+        return Column(ForeignKey(f'{CDM_SCHEMA}.visit_detail.visit_detail_id'))
 
     @declared_attr
     def measurement_source_value(cls):
@@ -512,6 +524,10 @@ class BaseMeasurementCdm531:
     @declared_attr
     def visit_occurrence(cls):
         return relationship('VisitOccurrence')
+
+    @declared_attr
+    def visit_detail(cls):
+        return relationship('VisitDetail')
 
 
 class BaseNoteCdm531:
@@ -568,7 +584,7 @@ class BaseNoteCdm531:
 
     @declared_attr
     def visit_detail_id(cls):
-        return Column(Integer)
+        return Column(ForeignKey(f'{CDM_SCHEMA}.visit_detail.visit_detail_id'))
 
     @declared_attr
     def note_source_value(cls):
@@ -601,6 +617,10 @@ class BaseNoteCdm531:
     @declared_attr
     def visit_occurrence(cls):
         return relationship('VisitOccurrence')
+
+    @declared_attr
+    def visit_detail(cls):
+        return relationship('VisitDetail')
 
 
 class BaseNoteNlpCdm531:
@@ -734,7 +754,7 @@ class BaseObservationCdm531:
 
     @declared_attr
     def visit_detail_id(cls):
-        return Column(Integer)
+        return Column(ForeignKey(f'{CDM_SCHEMA}.visit_detail.visit_detail_id'))
 
     @declared_attr
     def observation_source_value(cls):
@@ -787,6 +807,10 @@ class BaseObservationCdm531:
     @declared_attr
     def visit_occurrence(cls):
         return relationship('VisitOccurrence')
+
+    @declared_attr
+    def visit_detail(cls):
+        return relationship('VisitDetail')
 
 
 class BaseObservationPeriodCdm531:
@@ -981,7 +1005,7 @@ class BaseProcedureOccurrenceCdm531:
 
     @declared_attr
     def visit_detail_id(cls):
-        return Column(Integer)
+        return Column(ForeignKey(f'{CDM_SCHEMA}.visit_detail.visit_detail_id'))
 
     @declared_attr
     def procedure_source_value(cls):
@@ -1022,6 +1046,10 @@ class BaseProcedureOccurrenceCdm531:
     @declared_attr
     def visit_occurrence(cls):
         return relationship('VisitOccurrence')
+
+    @declared_attr
+    def visit_detail(cls):
+        return relationship('VisitDetail')
 
 
 class BaseSpecimenCdm531:
--- a/src/delphyne/cdm/cdm531/clinical_data.py
+++ b/src/delphyne/cdm/cdm531/clinical_data.py
@@ -1248,7 +1248,7 @@ class BaseVisitOccurrenceCdm531:
 
     @declared_attr
     def visit_concept_id(cls):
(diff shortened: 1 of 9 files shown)


src/delphyne/cdm/cdm531/clinical_data.py
```diff
@@ -388,6 +404,24 @@ def fact_id_2(cls):
...
+    def domain_concept_1(cls):
+        return relationship('Concept',
+                            primaryjoin='FactRelationship.domain_concept_id_1 == '
+                                        'Concept.concept_id')
+
+    @declared_attr
+    def domain_concept_2(cls):
+        return relationship('Concept',
+                            primaryjoin='FactRelationship.domain_concept_id_2 == '
+                                        'Concept.concept_id')
```
alepev, line 417: This is the same in both CDM 5 and 6, but: I think`domain_concept` fields should refer to the Domain table, not the Concept table (similar to what happens in Cost), so that in principle it's possible to validate that `fact_id` 1&2 belong to their respective domains. I will check the forums to see if it has been brought up before.

src/delphyne/cdm/cdm531/clinical_data.py
```diff
@@ -104,6 +104,10 @@ def provider(cls):
     def visit_occurrence(cls):
         return relationship('VisitOccurrence')
 
+    @declared_attr
+    def visit_detail(cls):
+        return relationship('VisitDetail')
+
 
 class BaseDeathCdm531:
```
alepev, line 112: The DDL for Death now looks like this:
```sql
CREATE TABLE @cdmDatabaseSchema.DEATH (
 
			person_id integer NULL, 
			death_date date NULL, 
			death_datetime TIMESTAMP NULL, 
			death_type_concept_id integer NULL, 
			cause_concept_id integer NULL, 
			cause_source_value varchar(50) NULL, 
			cause_source_concept_id integer NULL );  
```

I think something went really wrong when they rewrote this. I kept all the non-nullable constraints in our model.

src/delphyne/cdm/cdm531/derived_elements.py
```diff
@@ -46,11 +46,11 @@ def condition_concept_id(cls):
 
     @declared_attr
     def condition_era_start_date(cls):
-        return Column(Date, nullable=False)
+        return Column(DateTime, nullable=False)
 
     @declared_attr
     def condition_era_end_date(cls):
-        return Column(Date, nullable=False)
+        return Column(DateTime, nullable=False)
```
alepev, line 53: These variables were updated to `DateTime` for some reason (same in `DoseEra` and `DrugEra`). It doesn't make sense to me to call something date and make it datetime, but it's in their official documentation too, not just in the DDL: https://ohdsi.github.io/CommonDataModel/cdm531.html#Standardized_Derived_Elements

src/delphyne/cdm/cdm531/health_economics.py
```diff
@@ -99,11 +145,11 @@ def cost_event_id(cls):
 
     @declared_attr
     def cost_domain_id(cls):
-        return Column(String(20), nullable=False)
+        return Column(ForeignKey(f'{VOCAB_SCHEMA}.domain.domain_id'), nullable=False)
```
alepev, line 148: This should reflect the domain the `cost_event_id` referenced above belongs to.

src/delphyne/cdm/vocabularies/vocabularies.py
```diff
@@ -340,7 +340,7 @@ class BaseSourceToConceptMap:
 
     @declared_attr
     def source_code(cls):
-        return Column(Text, primary_key=True, index=True)
+        return Column(String(50), primary_key=True, index=True)
```
alepev, line 343: @Spayralbe might not like this! If there is not tool dependency on this constraint, I'm happy to revert to a more permissive format.. https://ohdsi.github.io/CommonDataModel/cdm531.html#SOURCE_TO_CONCEPT_MAP

alepev:
Now, something that requires some deep thinking:
- the previous official CDM v5.3.1 version contained the tables `Cohort` and `Cohort Attribute` as part of the standardized derived element, plus `Cohort Definition` and `Attribute Definition` as part of the vocabulary
- the fixed CDM v5.3.1 version contains only `Cohort Definition` and `Attribute Definition`, but, interestingly enough, the official documentation still refers to a mysterious `Cohort` table in the `Cohort Definition` description (see [official docs](https://ohdsi.github.io/CommonDataModel/cdm531.html#COHORT_DEFINITION)).
- our CDM v5.3.1 model contains `Cohort` (standardized derived elements) and `Cohort Definition` (vocabulary)

I was reasoning about the following:
- we should probably keep `Cohort` and ask someone involved with the CDM to check what happened to it
- I have no idea what to do about `Cohort Attribute`. We don't have it at the moment, so leave it out? Probably good to ask about that one as well.
- `Attribute Definition` is in the official CDM v5.3.1 model, but not in ours. It's a pain because it's part of the vocabulary tables, but only in CDM v5.3.1, not in v6.0. Supporting it means our vocabulary tables would become version-specific. Alternatively, we could put both `Attribute Definition` and `Cohort Definition` in a separate schema, either `cdm` or something else (I saw suggestions to relegate them to a `results` schema since they are user-populated for specific applications only). Not sure if that would break tools that strictly expect it to be in the `vocabulary` schema.. Are there any?

alepev:
..Other than that, in the "fixed" DDL and constraints files I discovered a few incorrect references, and several omissions. Rather underwhelming.

alepev:
Bonus question for @Spayralbe : what happens if there is a `Column` with a FK to another table, but no corresponding `relationship` defined?

src/delphyne/cdm/cdm531/clinical_data.py
Spayralbe: Agreed, let's keep those `NOT NULL` constraints.

src/delphyne/cdm/vocabularies/vocabularies.py
Spayralbe: 😄 Yes this one I actually changed on purpose to allow longer source code values. I'd prefer to keep it as `Text`.

Spayralbe:
> Bonus question for @Spayralbe : what happens if there is a `Column` with a FK to another table, but no corresponding `relationship` defined?

The earth will probably keep turning, but without a relationship attribute you cannot let SQLAlchemy automatically resolve PK - FK relationships when inserting new records (we rarely use it though). Also relationships can be used when creating a SQLAlchemy Query.

src/delphyne/cdm/cdm531/clinical_data.py
MaximMoinat: Actually, it can refer to a concept. And the documentation seems to very specific on a FK relation to the Concept table. 
For example the person domain: https://athena.ohdsi.org/search-terms/terms/1147314

src/delphyne/cdm/cdm531/health_economics.py
MaximMoinat: Agreed

src/delphyne/cdm/vocabularies/vocabularies.py
MaximMoinat: I have it from a good source that some people actually use way more permissive types on these text fields (e.g. also for `source_value` fields). The tools are just fine with that apparently.

src/delphyne/cdm/cdm531/derived_elements.py
MaximMoinat: Yeah, it is a bit inconsistent. But let's stick with the official definition.

MaximMoinat (approved):
Looks good to me!

Spayralbe (approved):

alepev:
> Now, something that requires some deep thinking:
> 
> * the previous official CDM v5.3.1 version contained the tables `Cohort` and `Cohort Attribute` as part of the standardized derived element, plus `Cohort Definition` and `Attribute Definition` as part of the vocabulary
> * the fixed CDM v5.3.1 version contains only `Cohort Definition` and `Attribute Definition`, but, interestingly enough, the official documentation still refers to a mysterious `Cohort` table in the `Cohort Definition` description (see [official docs](https://ohdsi.github.io/CommonDataModel/cdm531.html#COHORT_DEFINITION)).
> * our CDM v5.3.1 model contains `Cohort` (standardized derived elements). The vocabulary contain `Cohort Definition`. We don't have `Cohort Attribute` and `Attribute Definition`.
> 
> I was reasoning about the following:
> 
> * we should probably keep `Cohort` and ask someone involved with the CDM to check what happened to it
> * I have no idea what to do about `Cohort Attribute`. We don't have it at the moment, so leave it out? Probably good to ask about that one as well.
> * `Attribute Definition` is in the official CDM v5.3.1 model, but not in ours. It's a pain because it's part of the vocabulary tables, but only in CDM v5.3.1, not in v6.0. Supporting it means our vocabulary tables would become version-specific. Alternatively, we could put both `Attribute Definition` and `Cohort Definition` in a separate schema, either `cdm` or something else (I saw suggestions to relegate them to a `results` schema since they are user-populated for specific applications only). Not sure if that would break tools that strictly expect it to be in the `vocabulary` schema.. Are there any?

@Spayralbe @MaximMoinat any thoughts on this one? I'll investigate a little otherwise.. I could create a specific issue to close this one in the meantime

src/delphyne/cdm/vocabularies/vocabularies.py
alepev: Changed back to what it was, thanks!

src/delphyne/cdm/cdm531/clinical_data.py
alepev: OK, let's keep it like this. Could also be the other way around, i.e. why doesn't Cost refer to a `domain_concept` in the Concept table too? 🤷‍♀️ It's literally the only CDM table that references Domain.. bah!

Spayralbe:
@alepev IMO those tables can be removed. They are legacy tables that have moved to the results schema (used by ATLAS/webAPI) or have been removed completely. See https://github.com/OHDSI/CommonDataModel/issues/212

MaximMoinat:
I agree, we can remove these cohort tables. For most ETLs they are not relevant. FYI; it just so happens that in the clinical trial working group we have a convention of storing study information in the cohort definition table.

Spayralbe:
In that case we could also put those tables in a separate `legacy` folder where you could still import them from if you really want to, but we should remove them from the template-repo list of default tables.

alepev:
Thanks for the feedback! 

Translated into action:
1. I'll move both `Cohort` (cdm schema) and `Cohort Definition` (vocab schema) to a legacy folder in delphyne 
2. ..and since I'm already on it, add `Attribute Definition` and `Cohort Attribute` for full legacy support?
3. in terms of table binding, I will keep the tables attached to the appropriate schema as per official docs (vocab for `Cohort Definition` and `Attribute Definition`, cdm for `Cohort` and `Cohort Attribute`). 
4. in `delphyne-template`, I'll add a legacy section to the table imports file with the legacy tables commented out

Let me know if you disagree with any of it

alepev:
> I agree, we can remove these cohort tables. For most ETLs they are not relevant. FYI; it just so happens that in the clinical trial working group we have a convention of storing study information in the cohort definition table.

Question: in which schema do they keep this table usually? vocabulary as per docs, results, or other? would it make sense to make a different version of Cohort Definition as an extension specifically for clinical trials?

MaximMoinat:
We did not discuss which schema. And I think it really depends on the local implementation, could be vocab, results or the cdm schema. 

Let's not make things too complicated. I like the solution of making them legacy tables in schema's as specified by the documentation (although personally I would put them in the cdm schema)

Spayralbe:
> Thanks for the feedback!
> 
> Translated into action:
> 
>     1. I'll move both `Cohort` (cdm schema) and `Cohort Definition` (vocab schema) to a legacy folder in delphyne
> 
>     2. ..and since I'm already on it, add `Attribute Definition` and `Cohort Attribute` for full legacy support?
> 
>     3. in terms of table binding, should I keep the tables attached to the schema as per official docs (vocab for `Cohort Definition` and `Attribute Definition`, cdm for `Cohort` and `Cohort Attribute`)? Ideally a separate `results` schema would be better, but sounds an overkill just to support legacy stuff..
> 
>     4. in `delphyne-template`, I'll add a legacy section at the bottom of the table imports file with the legacy tables commented out
> 
> 
> Let me know if you disagree with any of it

Agreed! Only thing I'm wondering about is whether we should list those tables at all in the template (point 4), but it's fine with me either way.

alepev:
All done

alepev:
p.s. I put everything in the `cdm` schema, I also think it's a bit cleaner that way

Spayralbe:
Looks good 👍

PR #143, opened 2021-02-11
CodeMapper tests

### LINKED ISSUE #69: Add tests for CodeMapper
See changes in #66

alepev:
Fixes #69


alepev:
@Spayralbe I have the feeling the new tests could run a bit faster, I'm curious if you any suggestions about that

Spayralbe (commented):
Tests look good to me :)

Only thing I was wondering about is whether it's needed to use a for loop for testing the result properties, as compared to just writing out the individual assert statements directly.

Performance we can take up separately as part of #154

alepev:
Thanks, I will try the 

> Tests look good to me :)
> 
> Only thing I was wondering about is whether it's needed to use a for loop for testing the result properties, as compared to just writing out the individual assert statements directly.
> 
> Performance we can take up separately as part of #154

These constructs you mean?
```
    assert len(result_no_match) == 1
    for attr, value in expected_result_no_match.__dict__.items():
        assert getattr(result_no_match[0], attr) == value
```

It is a bit weird, but it does save typing :P I thought it will be handy if for some reason I have to add/remove some of the expected_result` object properties

alepev:
Oh, I see! I would say here it is pretty safe because I'm assigning values to the properties explicitly, but if I were getting it from somewhere else it can go really wrong.. There is a potential issue though if I create a new `CodeMapping` object and don't change the default property values, then there's no way of telling if the property is there or not.. I will add a test to check that `CodeMapping` instantiation behaves as it should, with all required properties and default values

PR #146, opened 2021-02-15
CodeMapper docs setup

### LINKED ISSUE #135: Document use of CodeMapper

alepev:
Closes #135

### DIFF
--- a/docs/code_mapper.rst
+++ b/docs/code_mapper.rst
@@ -15,7 +15,7 @@ For example, it is possible to automatically map ICD10CM ontology terms to their
 Once created, a mapping dictionary can be used in any transformation to quickly lookup target `concept_id` information
 for any source term.
 
-Note that creating a mapping dictionary is only possible for official (non-standard) OMOP vocabularies;
+Note that creating a mapping dictionary is only possible for official OMOP vocabularies;
 if a source vocabulary is not available from `Athena <https://athena.ohdsi.org/vocabulary/list>`_,
 other mapping methods should be considered (e.g. loading the mappings directly from file).
 
@@ -24,11 +24,11 @@ Creating a mapping dictionary
 
 Mapping dictionaries (in fact, :class:`.MappingDict` objects) can be created
 for one or multiple source vocabularies at once, identified by their OMOP `vocabulary_id`.
-When possible, it is recommended to use the `restrict_to_codes` argument to load mappings
+When possible, it is recommended to use the ``restrict_to_codes`` argument to load mappings
 only for the source codes that will be actually used in the transformations, thus limiting memory usage.
 
 `delphyne`'s Wrapper provides a ready-to-use :class:`.CodeMapper` instance under
-the attribute `code_mapper`. You can use it to create a mapping dictionary inside a transformation script as follows:
+the attribute ``code_mapper``. You can use it to create a mapping dictionary inside a transformation script as follows:
 
 .. code-block:: python
 
@@ -36,9 +36,10 @@ the attribute `code_mapper`. You can use it to create a mapping dictionary insid
                vocabulary_id='ICD10CM',
                restrict_to_codes=['R51', 'T68', 'B36.0'])
 
-If the same mapping dictionary is needed in several transformations, it might be more efficient to instantiate
-it as a wrapper attribute and reuse it, since the dictionary creation can take up considerable time.
-This is especially true for long `restrict_to_codes` lists, or when no `restrict_to_codes` argument is provided.
+.. note::
+   If the same mapping dictionary is needed in several transformations, it might be more efficient to instantiate
+   it as a Wrapper attribute and reuse it, since the dictionary creation can take up considerable time.
+   This is especially true for long ``restrict_to_codes`` lists, or when no ``restrict_to_codes`` argument is provided.
 
 Using a mapping dictionary
 ----------------------------
@@ -68,7 +69,7 @@ with the following attributes:
    mapping.target_vocabulary_id     # 'SNOMED'
 
 If a code is not found in the mapping dictionary, :meth:`~.MappingDict.lookup()` returns a list containing
-a single :class:`.CodeMapping` object with both `source_concept_id` and `target_concept_id` set to `0`.
+a single :class:`.CodeMapping` object with both `source_concept_id` and `target_concept_id` set to ``0``.
 
-Use the option `target_concept_id_only=True` to retrieve any matching `target_concept_id` (as integer) instead
-of full mapping objects. Use `first_only=True` to retrieve the first available match instead of a list of matches.
+Use the option ``target_concept_id_only=True`` to return `target_concept_id` instead of full mapping objects.
+Use ``first_only=True`` to retrieve the first available match instead of a list of all matches.
--- a/docs/code_mapper.rst
+++ b/docs/code_mapper.rst
@@ -8,10 +8,12 @@ Code mapper
 Mapping source codes to standard concept_ids
 --------------------------------------------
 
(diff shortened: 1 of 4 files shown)


docs/code_mapper.rst
```diff
@@ -0,0 +1,48 @@
...
+===================
+
+.. contents::
+    :local:
+    :backlinks: none
+
+Mapping source codes to standard concept_ids
+--------------------------------------------
+
+`delphyne`'s CodeMapper class enables the creation of mapping dictionaries from non-standard vocabulary terms to
```
alepev, line 11: How can I create a link from `CodeMapper` to the actual auto-generated documentation for CodeMapper?

docs/reference.rst
```diff
@@ -60,8 +60,8 @@ StcmLoader
 CodeMapper
 ----------
 
-.. autoclass:: src.delphyne.model.mapping.code_mapper.CodeMapper
-   :members:
+.. automodule:: src.delphyne.model.mapping.code_mapper
+   :members: CodeMapper, MappingDict, CodeMapping
```
alepev, line 64: I changed this to also be able to link to `MappingDict` and `CodeMapping` in the same module; alternatively, I could make these classes part of `CodeMapper` itself. (Note that I'm listing members explicitly because there is a module class, `Record`, that I don't want to appear in the documentation)

alepev:
Also big question: @Spayralbe how can I test that the docs build correctly? I get warnings for `pygments`, I guess I need that at least and perhaps a few other packages?

alepev:
@MaximMoinat there are some questions in this PR that can be interesting for you, since you also want to know about documentation

docs/code_mapper.rst
alepev: I will try using this, but still not sure how to see if the result is correct: https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects

docs/code_mapper.rst
```diff
@@ -0,0 +1,48 @@
...
+===================
+
+.. contents::
+    :local:
+    :backlinks: none
+
+Mapping source codes to standard concept_ids
+--------------------------------------------
+
+`delphyne`'s :class:`CodeMapper` class enables the creation of mapping dictionaries from non-standard vocabulary terms to
```
alepev, line 11: Still not sure if this `:class:CodeMapper` works. I get a warning _Unknown interpreted text role "class"_ even though this should be `sphinx` recognized syntax.

docs/code_mapper.rst
```diff
@@ -0,0 +1,48 @@
...
+Creating a mapping dictionary
+-----------------------------
+
+Mapping dictionaries (in fact, :class:`MappingDict` objects) can be created for one or multiple source vocabularies at once,
+identified by their OMOP `vocabulary_id`. When possible, it is recommended to use the `restrict_to_codes` argument
+to load mappings only for the source codes that will be actually used in the transformations, thus limiting memory usage.
+
+.. code-block:: python
+
+   mapping_dict = self.code_mapper.generate_code_mapping_dictionary(
```
alepev, line 29: I find it a bit confusing here to use `self`, it doesn't make clear this method is available in the wrapper. shall I replace with `wrapper`?

docs/code_mapper.rst
alepev: Despite the warnings, this works as expected:
```
:mod:`~src.delphyne.model.mapping.code_mapper.MappingDict`
```
The `~` at the beginning makes sure that only the last component (`MappingDict`) is shown in documentation.

docs/reference.rst
alepev: This doesn't work. To exclude `Record`, I have to explicitly use `:exclude-members`. Also, the order of classes specified in `:members:` has zero effect, they will still be shown in alphabetical order. I'll leave it like that..

docs/code_mapper.rst
alepev: It should be clearer now from the description

Spayralbe (commented):
Contents all look good to me 👍

For consistency: shall we use inline literals for any references to code (instead of cursive)? You can add them with double backticks in rst.

alepev:
> Contents all look good to me 👍
> 
> For consistency: shall we use inline literals for any references to code (instead of cursive)? You can add them with double backticks in rst.

Yes, I was wondering about the style indeed! Btw, I also wrote _delphyne_ in cursive and noticed you wrote it simply "Delphyne", I think I like the latter actually.

docs/code_mapper.rst
```diff
@@ -0,0 +1,75 @@
...
+===========
+
+.. contents::
+    :local:
+    :backlinks: none
+
+Mapping source codes to standard concept_ids
+--------------------------------------------
+
+`delphyne`'s :class:`.CodeMapper` class enables the creation
```
alepev, line 11: This line

docs/code_mapper.rst
```diff
@@ -0,0 +1,75 @@
...
+
+Creating a mapping dictionary
+-----------------------------
+
+Mapping dictionaries (in fact, :class:`.MappingDict` objects) can be created
+for one or multiple source vocabularies at once, identified by their OMOP `vocabulary_id`.
+When possible, it is recommended to use the `restrict_to_codes` argument to load mappings
+only for the source codes that will be actually used in the transformations, thus limiting memory usage.
+
+`delphyne`'s Wrapper provides a ready-to-use :class:`.CodeMapper` instance under
```
alepev, line 30: and this line. also mentioning "Wrapper" here, shall I keep it like that or make it a code reference (`wrapper`)?

Spayralbe: Might be nice to make it a code reference? Although I'd also have to revise the docs I've added :p

alepev: Agreed, and actually it should be `Wrapper`, capitalized 😅

alepev:
> > Contents all look good to me 👍
> > For consistency: shall we use inline literals for any references to code (instead of cursive)? You can add them with double backticks in rst.
> 
> Yes, I was wondering about the style indeed! Btw, I also wrote _delphyne_ in cursive and noticed you wrote it simply "Delphyne", I think I like the latter actually.

@Spayralbe btw about this one: I am also writing CDM table names in all caps (e.g. VOCABULARY) and CDM-specific entities (a.k.a. table columns) in cursive (e.g. _concept_id_). Do we want a convention for that too? Actually I think at the moment the only words in cursive are indeed those.

Spayralbe:
> > > Contents all look good to me 👍
> > > For consistency: shall we use inline literals for any references to code (instead of cursive)? You can add them with double backticks in rst.
> > 
> > 
> > Yes, I was wondering about the style indeed! Btw, I also wrote _delphyne_ in cursive and noticed you wrote it simply "Delphyne", I think I like the latter actually.
> 
> @Spayralbe btw about this one: I am also writing CDM table names in all caps (e.g. VOCABULARY) and CDM table columns in cursive (e.g. _concept_id_). Do we want a convention for that too? Actually I think at the moment the only words I have in cursive are those.. I wouldn't mind removing the cursive.

I was going to say, lets use the same conventions as used in the [CDM documentation](https://ohdsi.github.io/CommonDataModel), but there it also differs per page :p
Table names however seem to be consistently written in uppercase, so let's do that in our docs as well. 
For column names, shall we just go for lower case without any special markup?

alepev:
Agreed, I'll remove the formatting

alepev:
I also changed the section title to "Semantic mapping" and made "CodeMapper" a sub-section. I think that gives a better feeling to the reader of what this is about. Also provides a home for similar methods, for example once VariableConceptMapper will be implemented

PR #147, opened 2021-02-15
Add support for SQL Server

Spayralbe:
Part of #23 

Add SQL Server support
- Expand `MainConfig` and `Database` to be able to handle a different dialect than postgresql
- Minor table definition changes
- Dialect-specific method for standard vocabulary loading
- Changes in transformation metadata collection of raw SQL queries
- New way of checking for existing schemas when calling `create_schemas()`
- Additions to `ConstraintManager` to facilitate the reflection of check constraints when not supported by SQLAlchemy


alepev:
Do we have a SQL Server running somewhere for E2E testing? Or is it a lot of work to install locally if I want to try it out?

It would be interesting to run this after integrating the CodeMapper tests, I am afraid there might be some SQL dialect incompatibilities there

Spayralbe:
> Do we have a SQL Server running somewhere for E2E testing? Or is it a lot of work to install locally if I want to try it out?
> 
> It would be interesting to run this after integrating the CodeMapper tests, I am afraid there might be some SQL dialect incompatibilities there

It was a bit of a hassle, but I got it to work in a [Docker container](https://docs.microsoft.com/en-us/sql/linux/quickstart-install-connect-docker?view=sql-server-linux-ver15&preserve-view=true&pivots=cs1-bash).

src/delphyne/cdm/vocabularies/vocabularies.py
```diff
@@ -340,7 +342,7 @@ class BaseSourceToConceptMap:
 
     @declared_attr
     def source_code(cls):
-        return Column(Text, primary_key=True, index=True)
+        return Column(VARCHAR(1000), primary_key=True, index=True)
```
Spayralbe, line 345: For SQL Server, `Text` becomes `VARCHAR(MAX)` which it doesn't allow you to index. Hence the arbitrary large number I chose now.

src/delphyne/database/database.py
```diff
@@ -211,38 +224,48 @@ def tracked_session_scope(self,
...
+        existing_schemas = inspector.get_schema_names()
         for schema in self.schemas:
-            metadata.reflect(schema=schema)
+            if schema not in existing_schemas:
+                continue
+            existing_tables = inspector.get_table_names(schema=schema)
+            model_tables = self._model_tables[schema]
+            reflect_tables = [t for t in model_tables if t in existing_tables]
+            metadata.reflect(schema=schema, only=reflect_tables, resolve_fks=False)
         return metadata
```
Spayralbe, line 269: Updated this to make sure that any additional tables that may be present in the schemas used, but are not part of the ORM, will not be reflected.

docs/standard_vocab.rst
```diff
@@ -50,3 +47,12 @@ This will drop all indexes and constraints on the tables before insertion, and r
 
 To prevent accidental reloading of vocabularies, they can only be loaded if all the target tables are empty.
 If you do want to reload them, you therefore need to drop the contents manually.
+
+
+Limitations
+-----------
+
+As SQL Server's bulk insert method requires a local file path, it can only be used if you're running
+delphyne in the same environment as the database.
```
alepev, line 56: Meaning, with PostgreSQL you can instead have a source file locally, but load to a remote server? I think we did that at some point indeed.. almost forgot it was possible

src/delphyne/database/constraints/constraint_manager.py
```diff
@@ -59,6 +60,64 @@ def _create_constraint_lookup(metadata: MetaData) -> Dict[str, ConstraintOrIndex
...
+        # Check if SQLAlchemy supports check constraint reflection for
+        # the given DBMS
+        inspector = Inspector.from_engine(self._db.engine)
+        try:
+            inspector.get_check_constraints('foo')
+        except NotImplementedError:
+            return False
+        except NoSuchTableError:
+            return True
+        return True
```
alepev, line 118: would this ever return True without encountering a `NoSuchTableError`? just curious

alepev: ..is that even a real error? 😅

src/delphyne/model/standard_vocab/standard_vocab_loader.py
```diff
@@ -118,6 +124,35 @@ def _insert_vocab_file(self, table: str, vocab_file: Path) -> None:
...
+        FROM '{vocab_file}'
+        WITH (
+        KEEPIDENTITY,
+        FIRSTROW = 2,
+        FIELDTERMINATOR = '\t',
+        ROWTERMINATOR = '0x0a',
+        ERRORFILE = '{error_file}',
+        TABLOCK
+        );
+        """
```
alepev, line 143: I did not realize how ugly this was 🙈 but good you managed to differentiate while keeping the same `_insert_vocab_file()` interface

docs/standard_vocab.rst
Spayralbe: You can yes. Although it would be significantly slower as you also need to upload the data to the server.

src/delphyne/database/constraints/constraint_manager.py
Spayralbe: If there happens to be a table named `foo` :p

src/delphyne/model/standard_vocab/standard_vocab_loader.py
Spayralbe: Yeah, it's almost a direct copy of the query provided by the cdm repo.

src/delphyne/database/constraints/constraint_manager.py
```diff
@@ -116,16 +175,34 @@ class ConstraintManager:
...
+        # check constraint reflection for the DBMS
+        meta = self._db.reflected_metadata
+        if self._chk_constraints.chk_support:
+            return meta
+        for chk_constraint in self._chk_constraints.all_chk_constraints:
+            table_name = f'{chk_constraint.schema_name}.{chk_constraint.table_name}'
+            if table_name not in meta.tables:
+                continue
+            table: Table = meta.tables.get(table_name)
+            dummy_constraint = CheckConstraint('', name=chk_constraint.chk_name, table=table)
```
alepev, line 193: I'm not sure how this dummy will apply in practice when used in the lookups below. it doesn't do anything right (`sqltext` is emtpy..)? Did you add this because SQL Server does not support constraints?

src/delphyne/database/database.py
```diff
@@ -76,27 +86,30 @@ def from_config(cls, config: MainConfig, base) -> Database:
...
+        )
+        if not password and not Database._can_connect_without_password(url):
+            url.password = getpass('Database password:')
+        return cls(uri=url, schema_translate_map=config.schema_translate_map, base=base)
 
     @staticmethod
-    def _password_needed(uri: str) -> bool:
+    def _can_connect_without_password(uri: URL) -> bool:
+        logger.info('Attempting to connect without password')
         logger.disabled = True
```
alepev, line 106: this name is better indeed. 
curiosity: what happens if you don't disable the logger? is it that verbose?

src/delphyne/database/constraints/constraint_manager.py
```diff
@@ -521,6 +599,10 @@ def _add_constraint_in_db(self,
         assert errors in _VALID_ERRORS_OPTIONS
```
alepev, line 599: I didn't notice this before. I thought assert was a test-specific thing.. what if this evaluates to false? it will interrupt execution with an exception right?

alepev (approved):
OK, I find the ConstraintManager still rather complex, but as far as I can follow I can understand which methods are called for Postgres vs SQL Server.. good luck with testing! 😅

src/delphyne/database/constraints/constraint_manager.py
alepev: This is one of the hardest parts for me. The code does find constraints (`for chk_constraint in self._chk_constraints.all_chk_constraints`...), but in the reflected metadata they will all lack that first `CheckConstraint` argument.. I'm not sure how it is handled by methods that add / drop constraints

Spayralbe: These dummy `CheckConstraints` are only added so they can be used in the `DropConstraint` call. `sqltext` remains empty, as it is irrelevant for this call. In the end what matters is the `ALTER TABLE DROP CONSTRAINT` call that is compiled by SQLAlchemy, and for that it's only relevant on what table the constraint is active and what the constraint name is.

src/delphyne/database/database.py
Spayralbe: Unwanted third-party library logging indeed

src/delphyne/database/constraints/constraint_manager.py
Spayralbe: Yes, it will raise an assertionerror when you provide an unknown option. I ~stole this idea~ was inspired by the way pandas does this when changing column dtypes.

Issue #126, opened 2021-01-20
Provide examples of transformation strategies for different performance scenarios


alepev:
Provide examples of how to query other tables in transformations, other tricks to reduce performance issues (things that are up to the user how to implement)

alepev:
+ refer the user to documentation about alternative transformation wrappers (#145)

alepev:
Include the following:
- explain the bulk_mode option in the SQLAlchemy ORM transformations (both normal and batch), with caveats
- explain that main difference between transformations styles is that Raw SQL and SQLAlchemy queries are ideal for table-to-table transformations, while SQLAlchemy ORM can be used for this but it will be significantly slower. it should be used primarily if reading data from file, or if you have complex processing steps that are difficult to implement otherwise

Spayralbe:
Some suggestions on performance considerations are already included. 
For now we decided to only add a new section at the end of the Transformations doc page, which explains the pros and cons of the different transformation options.

PR #155, opened 2021-02-19
Short transformations guide

### LINKED ISSUE #145: Document use of different transformation wrappers
e.g. `execute_transformation` vs `execute_batch_transformation`, and examples of how transformation itself should be adapted for each

MaximMoinat:
Closes #145 

Have not tested the rst syntax.

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -1,5 +1,0 @@
-
-### Building the docs
-Edit the .rst, then execute `nox -s docs` in the main projeect folder.
-
-Then open `docs/_build/index.html` in your browser to preview the docs page.
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -1,0 +1,5 @@
+
+### Building the docs
+Edit the .rst, then execute `nox -s docs` in the main projeect folder.
+
+Then open `docs/_build/index.html` in your browser to preview the docs page.
--- a/docs/transformations.rst
+++ b/docs/transformations.rst
@@ -69,13 +69,14 @@ The wrapper commits the yielded records to the database in batches of 10_000 rec
             )
             yield record
 
-In the wrapper's run method, these two transformations can be called like this.
+Inside a wrapper method, these two transformations can be called like this.
 Specifying the batch size is optional.
 
 .. code-block:: python
-
-    self.execute_transformation(my_transformation)
-    self.execute_batch_transformation(my_batch_transformation, batch_size=10000)
+    def run(self):
+        ...
+        self.execute_transformation(my_transformation)
+        self.execute_batch_transformation(my_batch_transformation, batch_size=10000)
 
 
 Raw SQL
@@ -104,12 +105,11 @@ The convention is to put these transformation scripts in the `src/main/sql` fold
      ...
     FROM @source_schema.<source_table>
 
-Executing the file is then done by adding the following line to the wrapper's run method.
+Executing the file is then done by adding the following line to a wrapper method.
 Only the filename has to be provided, delphyne will look for the file in the `src/main/sql` folder.
 
 .. code-block:: python
-
-    execute_sql_file('my_file.sql')
+        self.execute_sql_file('my_file.sql')
 
 
 SQLAlchemy query
--- a/docs/transformations.rst
+++ b/docs/transformations.rst
@@ -7,71 +7,106 @@ Transformations
 
 General intro
 -------------
-By extending the :mod:`~src.delphyne.wrapper.Wrapper` a number of transformation functions are available.
+The delphyne :class:`.Wrapper` offers different ways to execute transformation scripts.
 Basically, there are two options:
- - Write a python function, returning the results as ORM objects.
+ - Write a Python function, returning the results as ORM objects.
  - Write a SQL query, directly inserting into the target table.
 
-For both options the :mod:`~src.delphyne.wrapper.Wrapper` has build-in methods that coordinate the execution.
-These methods, detailed below, use SqlAlchemy to handle database operations and log the execution statistics.
-This provides a standardised interface independent of whether you are using Python or Sql for the transformation logic.
+For both options the :class:`.Wrapper` has built-in methods that coordinate the execution.
+These methods, detailed below, use SQLAlchemy to handle database operations and log the execution statistics.
+This provides a standardised interface independent of whether you are using Python or SQL for the transformation logic.
+
+It is good practice to create a separate transformation file (either Python or SQL) for each table to table
+transformation.
+This improves maintainability and allows for delphyne creating a detailed mapping log.
+In addition, the advised convention for the transformation file naming is `<source_table>_to_<target_table>.py/sql`.
 
 ORM
 ---
-The ORM approach allows the use of pure python to write transformations.
-This function takes the wrapper as input and produces OMOP records as SqlAlchemy ORM objects.
+The ORM approach allows the use of pure Python to write transformations.
+This function takes the wrapper as input and produces OMOP records as SQLAlchemy ORM objects.
 It can return these objects in two different ways (see below).
-While records from multiple different OMOP tables can be created in the same function,
- it is good practice to create a separate file for each table to table transformation.
-See delphyne-template for examples.
+The convention is to put the transformation scripts in the `src/main/python/transformation` folder.
+`See delphyne-template for examples. <https://github.com/thehyve/delphyne-template/tree/master/src/main/python/transformation>`_
 
 There are two ways of execution the transformation, each requiring a different return type.
-The first mode, used by :mod:`~src.delphyne.wrapper.Wrapper.execute_transformation`.
+The first mode, used by :meth:`.Wrapper.execute_transformation`.
 This expects the transformation function to return the full list of ORM objects at once.
 All returned records are committed to the database at the same time.
 
 While this is straightforward, it does require to store all records in memory.
-For large tables there is a 'batch' mode :mod:`~src.delphyne.wrapper.Wrapper.execute_batch_transformation`.
+For large tables there is a 'batch' mode :meth:`.Wrapper.execute_batch_transformation`.
 This expects the transformation function to yield the ORM objects one at a time.
 The wrapper will commits the yielded records to the database in batches of given size.
 
 The pseudo-code for a transformation is given below, with the only difference between the two
 execution modes being the way the omop records are returned.
-```python
-def my_transformation(wrapper):
-    # read source data
-    for row in source:
-        # transformation logic...
-        omop_record = OmopTable(  # e.g. Person, Observation, ...
-            person_id=...,
-            concept_id=...,
-            ...
-        )
-        # yield OR store in list
-    # return list (if not yielding)
-```
+
+Below we give two Python pseudo-codes, one for a 'regular' transformation and a batch transformation.
+
+**Regular**
+
+.. code-block:: python
+
+    def my_transformation(wrapper):
+        # ...read source data...
+        records_to_insert = []
+        for row in source:
+            # ...transformation logic...
+            omop_record = wrapper.cdm.Observation(
+                person_id=...,
+                observation_concept_id=...,
+                ...
+            )
+            records_to_insert.append(records_to_insert)
+        return records_to_insert
+
+**Batch**
+
+.. code-block:: python
+
+    def my_batch_transformation(wrapper):
+        # ...read source data...
+        for row in source:
+            # ...transformation logic...
+            omop_record = wrapper.cdm.Observation(
+                person_id=...,
+                observation_concept_id=...,
+                ...
+            )
+            yield omop_record
+
+In the wrapper these can be executed like this:
+
+.. code-block:: python
+
+    self.execute_transformation(my_transformation)
+    self.execute_batch_transformation(my_batch_transformation)
+
 
 
 Raw SQL
 -------------
-SQL queries can easily be executed with through the wrapper.
-In case of just executing a simple query, the method :mod:`~src.delphyne.wrapper.Wrapper.execute_sql_query` is used.
-If the sql query is saved in a file, the method :mod:`~src.delphyne.wrapper.Wrapper.execute_sql_file` is used.
+SQL queries can easily be executed with the wrapper.
+In case of just executing a simple query, the method :meth:`.Wrapper.execute_sql_query` is used.
+If the SQL query is saved in a file, the method :meth:`.Wrapper.execute_sql_file` is used.
 
-The sql query should handle the insertion of records.
+The SQL query should handle the insertion of records.
 The easiest way to create a transformations with sql is by following the template given here.
 If you have defined your transformation in Rabbit-in-a-Hat, then
- you can directly export `a sql skeleton in this format <http://ohdsi.github.io/WhiteRabbit/RabbitInAHat.html#generating_a_sql_skeleton_(v090)>`_.
+you can directly export `a SQL skeleton in this format <http://ohdsi.github.io/WhiteRabbit/RabbitInAHat.html#generating_a_sql_skeleton_(v090)>`_.
+The convention is to put these transformation scripts in the `src/main/sql` folder.
+`See delphyne-template for examples. <https://github.com/thehyve/delphyne-template/tree/master/src/main/sql>`_
 
-```sql
-INSERT INTO @cdmSchma.<target_table> (
- <target_column1>,
- <target_column2>,
- ...
-)
-SELECT
- <source_column1>,
- <source_column2>,
- ...
-FROM @sourceSchema.<source_table>
-```
+.. code-block:: sql
+
+    INSERT INTO @cdm_schema.<target_table> (
+     <target_column1>,
+     <target_column2>,
+     ...
+    )
+    SELECT
+     <source_column1>,
+     <source_column2>,
+     ...
+    FROM @source_schema.<source_table>
--- a/docs/transformations.rst
+++ b/docs/transformations.rst
@@ -73,6 +73,7 @@ Inside a wrapper method, these two transformations can be called like this.
 Specifying the batch size is optional.
 
 .. code-block:: python
+
     def run(self):
         ...
         self.execute_transformation(my_transformation)
@@ -109,6 +110,9 @@ Executing the file is then done by adding the following line to a wrapper method
 Only the filename has to be provided, delphyne will look for the file in the `src/main/sql` folder.
 
 .. code-block:: python
+
+    def run(self):
+        ...
         self.execute_sql_file('my_file.sql')
 
 
@@ -143,11 +147,13 @@ Relational Database Management System (RDBMS).
 In case the source table is not part of the CDM schema, you can obtain it with the following method, which leverages SQLAlchemy's ability to create reflected table objects from the database itself:
 
 .. code-block:: python
-        source_table = wrapper.get_table(schema='my_source_schema', table_name='my_source_table')
+
+    source_table = wrapper.get_table(schema='my_source_schema', table_name='my_source_table')
         
 Inside a wrapper method, the transformations can be called like this, similar to ORM transformations.
 
 .. code-block:: python
+
     def run(self):
         ...
         self.execute_sql_transformation(my_sql_transformation)
--- a/docs/transformations.rst
+++ b/docs/transformations.rst
@@ -6,11 +6,72 @@ Transformations
     :backlinks: none
 
 General intro
-
-
-Raw SQL files
 -------------
+By extending the :mod:`~src.delphyne.wrapper.Wrapper` a number of transformation functions are available.
+Basically, there are two options:
+ - Write a python function, returning the results as ORM objects.
+ - Write a SQL query, directly inserting into the target table.
 
+For both options the :mod:`~src.delphyne.wrapper.Wrapper` has build-in methods that coordinate the execution.
+These methods, detailed below, use SqlAlchemy to handle database operations and log the execution statistics.
+This provides a standardised interface independent of whether you are using Python or Sql for the transformation logic.
 
 ORM
 ---
+The ORM approach allows the use of pure python to write transformations.
+This function takes the wrapper as input and produces OMOP records as SqlAlchemy ORM objects.
+It can return these objects in two different ways (see below).
+While records from multiple different OMOP tables can be created in the same function,
+ it is good practice to create a separate file for each table to table transformation.
+See delphyne-template for examples.
+
+There are two ways of execution the transformation, each requiring a different return type.
+The first mode, used by :mod:`~src.delphyne.wrapper.Wrapper.execute_transformation`.
+This expects the transformation function to return the full list of ORM objects at once.
+All returned records are committed to the database at the same time.
+
+While this is straightforward, it does require to store all records in memory.
+For large tables there is a 'batch' mode :mod:`~src.delphyne.wrapper.Wrapper.execute_batch_transformation`.
+This expects the transformation function to yield the ORM objects one at a time.
+The wrapper will commits the yielded records to the database in batches of given size.
+
+The pseudo-code for a transformation is given below, with the only difference between the two
+execution modes being the way the omop records are returned.
+```python
+def my_transformation(wrapper):
+    # read source data
+    for row in source:
+        # transformation logic...
+        omop_record = OmopTable(  # e.g. Person, Observation, ...
+            person_id=...,
+            concept_id=...,
+            ...
+        )
+        # yield OR store in list
+    # return list (if not yielding)
+```
+
+
+Raw SQL
+-------------
+SQL queries can easily be executed with through the wrapper.
+In case of just executing a simple query, the method :mod:`~src.delphyne.wrapper.Wrapper.execute_sql_query` is used.
+If the sql query is saved in a file, the method :mod:`~src.delphyne.wrapper.Wrapper.execute_sql_file` is used.
+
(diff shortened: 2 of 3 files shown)


docs/transformations.rst
```diff
@@ -6,11 +6,72 @@ Transformations
     :backlinks: none
 
 General intro
-
-
-Raw SQL files
 -------------
+By extending the :mod:`~src.delphyne.wrapper.Wrapper` a number of transformation functions are available.
```
alepev, line 10: you can replace this and other occurrences with :class:`.Wrapper`, it will work :)

docs/transformations.rst
```diff
@@ -6,11 +6,72 @@ Transformations
...
 ---
+The ORM approach allows the use of pure python to write transformations.
+This function takes the wrapper as input and produces OMOP records as SqlAlchemy ORM objects.
+It can return these objects in two different ways (see below).
+While records from multiple different OMOP tables can be created in the same function,
+ it is good practice to create a separate file for each table to table transformation.
+See delphyne-template for examples.
+
+There are two ways of execution the transformation, each requiring a different return type.
+The first mode, used by :mod:`~src.delphyne.wrapper.Wrapper.execute_transformation`.
```
alepev, line 29: ..and this with :meth:`~.Wrapper.execute_transformation()`

docs/transformations.rst
alepev: btw: "by extending the Wrapper" sounds a bit weird. I would say that the Wrapper offers different ways to execute transformation scripts?

docs/transformations.rst
```diff
@@ -6,11 +6,72 @@ Transformations
...
-
-
-Raw SQL files
 -------------
+By extending the :mod:`~src.delphyne.wrapper.Wrapper` a number of transformation functions are available.
+Basically, there are two options:
+ - Write a python function, returning the results as ORM objects.
+ - Write a SQL query, directly inserting into the target table.
 
+For both options the :mod:`~src.delphyne.wrapper.Wrapper` has build-in methods that coordinate the execution.
```
alepev, line 15: builT-in (if you keep this sentence, maybe not needed)

docs/transformations.rst
```diff
@@ -6,11 +6,72 @@ Transformations
...
+These methods, detailed below, use SqlAlchemy to handle database operations and log the execution statistics.
```
alepev, line 16: the capitalization is "SQLAlchemy"

alepev: Same for "SQL" below

docs/transformations.rst
```diff
@@ -6,11 +6,72 @@ Transformations
...
+This provides a standardised interface independent of whether you are using Python or Sql for the transformation logic.
 
 ORM
 ---
+The ORM approach allows the use of pure python to write transformations.
+This function takes the wrapper as input and produces OMOP records as SqlAlchemy ORM objects.
+It can return these objects in two different ways (see below).
+While records from multiple different OMOP tables can be created in the same function,
+ it is good practice to create a separate file for each table to table transformation.
+See delphyne-template for examples.
```
alepev, line 26: It could be useful to link to the template repository here, specifically the `transformations` folder

docs/transformations.rst
alepev: (Check other occurrences too, it's a minor thing really but should be consistent). 
Another one I noticed: capitalization of Python (sometimes written lowercase, sometimes uppercase)

docs/transformations.rst
alepev: Actually, it might be good to add specific instructions on where to place the transformations in the template:
- for ORM-based: `src/main/python/transformation`
- for SQL-based: `src/main/sql`

docs/transformations.rst
```diff
@@ -6,11 +6,72 @@ Transformations
...
+These methods, detailed below, use SqlAlchemy to handle database operations and log the execution statistics.
+This provides a standardised interface independent of whether you are using Python or Sql for the transformation logic.
 
 ORM
 ---
+The ORM approach allows the use of pure python to write transformations.
+This function takes the wrapper as input and produces OMOP records as SqlAlchemy ORM objects.
+It can return these objects in two different ways (see below).
+While records from multiple different OMOP tables can be created in the same function,
+ it is good practice to create a separate file for each table to table transformation.
```
alepev, line 25: This part is true for all transformations, move at the top before going into the details of ORM vs SQL

docs/transformations.rst
alepev: it works irrespectively, it's just shorter if you need to edit it later in the raw .rst file

docs/transformations.rst
```diff
@@ -6,11 +6,72 @@ Transformations
...
+    for row in source:
+        # transformation logic...
+        omop_record = OmopTable(  # e.g. Person, Observation, ...
+            person_id=...,
+            concept_id=...,
+            ...
+        )
+        # yield OR store in list
+    # return list (if not yielding)
+```
```
alepev, line 52: I suggest providing separate explicit examples for the case with return and yield, and add blocks showing how you would execute the corresponding transformation with either `execute_transformation` or `execute_batch_transformation` in the Wrapper. I think you can use of the the OMOP tables explicitly, say `Person`

docs/transformations.rst
```diff
@@ -6,11 +6,72 @@ Transformations
...
+
+While this is straightforward, it does require to store all records in memory.
+For large tables there is a 'batch' mode :mod:`~src.delphyne.wrapper.Wrapper.execute_batch_transformation`.
+This expects the transformation function to yield the ORM objects one at a time.
+The wrapper will commits the yielded records to the database in batches of given size.
+
+The pseudo-code for a transformation is given below, with the only difference between the two
+execution modes being the way the omop records are returned.
+```python
+def my_transformation(wrapper):
```
alepev, line 41: You need an empty line between the text and the start of the code block, otherwise it doesn't display correctly. Also between `python` and the first line of code.

docs/transformations.rst
```diff
@@ -6,11 +6,72 @@ Transformations
...
+ <target_column1>,
+ <target_column2>,
+ ...
+)
+SELECT
+ <source_column1>,
+ <source_column2>,
+ ...
+FROM @sourceSchema.<source_table>
+```
```
alepev, line 77: same here, you will need something like `.. code-block:: sql` (see if it works, I haven't tried that one)

docs/transformations.rst
```diff
@@ -6,11 +6,72 @@ Transformations
...
+In case of just executing a simple query, the method :mod:`~src.delphyne.wrapper.Wrapper.execute_sql_query` is used.
+If the sql query is saved in a file, the method :mod:`~src.delphyne.wrapper.Wrapper.execute_sql_file` is used.
+
+The sql query should handle the insertion of records.
+The easiest way to create a transformations with sql is by following the template given here.
+If you have defined your transformation in Rabbit-in-a-Hat, then
+ you can directly export `a sql skeleton in this format <http://ohdsi.github.io/WhiteRabbit/RabbitInAHat.html#generating_a_sql_skeleton_(v090)>`_.
+
+```sql
+INSERT INTO @cdmSchma.<target_table> (
```
alepev, line 67: typo in `@cdmSchema`

docs/transformations.rst
```diff
@@ -6,11 +6,72 @@ Transformations
...
+            ...
+        )
+        # yield OR store in list
+    # return list (if not yielding)
+```
+
+
+Raw SQL
+-------------
+SQL queries can easily be executed with through the wrapper.
```
alepev, line 57: "with" or "through"

docs/transformations.rst
alepev: actually, the placeholder name should be `@cdm_schema`. this will be taken straight from the `schema_translate_map` section in the configuration file

docs/transformations.rst
```diff
@@ -6,11 +6,72 @@ Transformations
...
+INSERT INTO @cdmSchma.<target_table> (
+ <target_column1>,
+ <target_column2>,
+ ...
+)
+SELECT
+ <source_column1>,
+ <source_column2>,
+ ...
+FROM @sourceSchema.<source_table>
```
alepev, line 76: I was wondering if we could give an example where the data is read from file rather than from another schema, since that's our main use case.

alepev (commented):
Good start, needs some fixes. Btw, if you want to see the result, run `nox -s docs`, then head to `docs/_build` and open `index.html`. IntelliJ should show you some browser icons in the upper right corner. (Otherwise: right-click > Open in > Browser)

docs/transformations.rst
```diff
@@ -6,11 +6,72 @@ Transformations
...
+        # yield OR store in list
+    # return list (if not yielding)
+```
+
+
+Raw SQL
+-------------
+SQL queries can easily be executed with through the wrapper.
+In case of just executing a simple query, the method :mod:`~src.delphyne.wrapper.Wrapper.execute_sql_query` is used.
+If the sql query is saved in a file, the method :mod:`~src.delphyne.wrapper.Wrapper.execute_sql_file` is used.
```
alepev, line 59: Also in this case, could be good to write out an example. There is a difference with `execute_transformation()`, that is the full SQL file name has to written out explicitly as a string (e.g. `execute_sql_file('my_file.sql'`))

docs/transformations.rst
MaximMoinat: Well, extending this class is what you do right?
```python
class my_wrapper(delphyne.Wrapper):
    ...
```

MaximMoinat: Will still use your suggestion :)

docs/transformations.rst
MaximMoinat: Cool, that is actually very helpful to learn a bit on rst! :D

docs/transformations.rst
MaximMoinat: For pure SQL, that is not our main use case, right?

MaximMoinat:
Thanks bunches @alepev! I updated the text accordingly

docs/transformations.rst
```diff
@@ -5,12 +5,111 @@ Transformations
...
 
+Introduction
+-------------
+The delphyne :class:`.Wrapper` offers different ways to execute transformation scripts.
+Basically, there are three options:
+ - Write a Python function, returning the results as `SQLAlchemy ORM objects <https://docs.sqlalchemy.org/en/14/orm/tutorial.html>`_.
+ - Write a SQL query, directly inserting into the target table.
+ - Write a Python function, creating a query using the `SQLAlchemy expression language <https://docs.sqlalchemy.org/en/14/core/tutorial.html>`_.
+
+For both options the :class:`.Wrapper` has built-in methods that coordinate the execution.
```
alepev, line 17:
```suggestion
For all of these options, the :class:`.Wrapper` has built-in methods that coordinate the execution.
```

alepev: (cannot use both with > 2)

docs/transformations.rst
```diff
@@ -5,12 +5,111 @@ Transformations
...
+transformation.
+This improves maintainability and allows for delphyne creating a detailed mapping log.
+In addition, the advised convention for the transformation file naming is ``<source_table>_to_<target_table>.py/sql``.
+
+
+SQLAlchemy ORM
+---------------
+The SQLAlchemy ORM approach allows the use of 'simple' Python to write transformations.
+This function takes the wrapper as input and produces OMOP records as SQLAlchemy ORM objects.
+It can return these objects in two different ways (see below).
```
alepev, line 31: I think you can remove this line, it's says again below that there's 2 ways of executing

docs/transformations.rst
```diff
@@ -5,12 +5,111 @@ Transformations
...
+The convention is to put the transformation scripts in the `src/main/python/transformation` folder.
+`See delphyne-template for examples. <https://github.com/thehyve/delphyne-template/tree/master/src/main/python/transformation>`_
+
+There are two ways of execution the transformation, each requiring a different return type.
+The first mode, used by :meth:`.Wrapper.execute_transformation`.
+This expects the transformation function to return the full list of ORM objects at once.
```
alepev, line 37: The first sentence ends strangely with a dot, I think you can merge them ("[...] Wrapper.execute_transformation expects the transformation function..")

docs/transformations.rst
```diff
@@ -5,12 +5,111 @@ Transformations
...
+All returned records are committed to the database at the same time.
+
+.. code-block:: python
+
+    def my_transformation(wrapper):
+        # ...read source data...
+        records_to_insert = []
+        for row in source:
+            # ...transformation logic...
+            omop_record = wrapper.cdm.Observation(
```
alepev, line 47: not sure about that "omop_record".. just "record" or "cdm_record"? :P

docs/transformations.rst
```diff
@@ -5,12 +5,111 @@ Transformations
...
+                person_id=...,
+                observation_concept_id=...,
+                ...
+            )
+            yield omop_record
+
+In the wrapper these two transformations can be called like this:
```
alepev, line 73: Suggestion: in other places we specifically say to add calls to the Wrapper's **run method**, so you could change it to make it very explicit

docs/transformations.rst
```diff
@@ -5,12 +5,111 @@ Transformations
...
+
+    self.execute_transformation(my_transformation)
+    self.execute_batch_transformation(my_batch_transformation)
+
+
+Raw SQL
 -------------
+SQL queries can easily be executed with the wrapper.
+In case of just executing a simple query, the method :meth:`.Wrapper.execute_sql_query` is used.
+If the SQL query is saved in a file, the method :meth:`.Wrapper.execute_sql_file` is used.
```
alepev, line 85: Also for these, please provide a code block at the end of the paragraph showing how these calls would appear in the wrapper method

docs/transformations.rst
```diff
@@ -5,12 +5,111 @@ Transformations
...
 
+The SQL query should handle the insertion of records.
+The easiest way to create a transformations with sql is by following the template given here.
```
alepev, line 88: sql -> SQL (check the file to see if it happens anywhere else)

docs/transformations.rst
```diff
@@ -5,12 +5,111 @@ Transformations
...
+    )
+    SELECT
+     <source_column1>,
+     <source_column2>,
+     ...
+    FROM @source_schema.<source_table>
+
+
+SQLAlchemy query
+-------------
```
alepev, line 109: the dashes should be the same length as the title

docs/transformations.rst
```diff
@@ -5,12 +5,111 @@ Transformations
...
+In the wrapper these two transformations can be called like this:
+
+.. code-block:: python
+
+    self.execute_transformation(my_transformation)
+    self.execute_batch_transformation(my_batch_transformation)
+
+
+Raw SQL
 -------------
```
alepev, line 82: The dashed should be the same length as the title

docs/transformations.rst
```diff
@@ -5,12 +5,111 @@ Transformations
...
+This provides a standardised interface independent of whether you are using Python or SQL for the transformation logic.
+
+It is good practice to create a separate transformation file (either Python or SQL) for each table to table
+transformation.
+This improves maintainability and allows for delphyne creating a detailed mapping log.
+In addition, the advised convention for the transformation file naming is ``<source_table>_to_<target_table>.py/sql``.
+
+
+SQLAlchemy ORM
+---------------
```
alepev, line 28: The dashes should be the same length as the title

docs/transformations.rst
```diff
@@ -5,12 +5,111 @@ Transformations
     :local:
     :backlinks: none
 
-General intro
 
+Introduction
+-------------
```
alepev, line 10: The dashes should be the same length as the title

docs/transformations.rst
```diff
@@ -5,12 +5,111 @@ Transformations
...
+The delphyne :class:`.Wrapper` offers different ways to execute transformation scripts.
+Basically, there are three options:
+ - Write a Python function, returning the results as `SQLAlchemy ORM objects <https://docs.sqlalchemy.org/en/14/orm/tutorial.html>`_.
+ - Write a SQL query, directly inserting into the target table.
+ - Write a Python function, creating a query using the `SQLAlchemy expression language <https://docs.sqlalchemy.org/en/14/core/tutorial.html>`_.
```
alepev, line 15:
```suggestion
    - Write a Python function, returning the results as `SQLAlchemy ORM objects <https://docs.sqlalchemy.org/en/14/orm/tutorial.html>`_.
    - Write a SQL query, directly inserting into the target table.
    - Write a Python function, creating a query using the `SQLAlchemy expression language <https://docs.sqlalchemy.org/en/14/core/tutorial.html>`_.
```

alepev: This worked fine, but somehow when generating the docs with sphinx I still get a complaint on the indentation level. I made it 4 spaces and the warning is gone.

alepev, line 15: when generating the docs with sphinx I get a complaint, add an empty line before the first bullet point and it will go away

alepev: Another note: should the links point one step higher, to the ORM/core reference instead of directly to the tutorial?

docs/transformations.rst
```diff
@@ -5,12 +5,111 @@ Transformations
...
+
+SQLAlchemy ORM
+---------------
+The SQLAlchemy ORM approach allows the use of 'simple' Python to write transformations.
+This function takes the wrapper as input and produces OMOP records as SQLAlchemy ORM objects.
+It can return these objects in two different ways (see below).
+The convention is to put the transformation scripts in the `src/main/python/transformation` folder.
+`See delphyne-template for examples. <https://github.com/thehyve/delphyne-template/tree/master/src/main/python/transformation>`_
+
+There are two ways of execution the transformation, each requiring a different return type.
```
alepev, line 35: "Two ways of executiNG" ?

docs/transformations.rst
```diff
@@ -5,12 +5,111 @@ Transformations
...
+                observation_concept_id=...,
+                ...
+            )
+            records_to_insert.append(records_to_insert)
+        return records_to_insert
+
+While this is straightforward, it does require to store all records in memory.
+For large tables there is a 'batch' mode :meth:`.Wrapper.execute_batch_transformation`.
+This expects the transformation function to yield the ORM objects one at a time.
+The wrapper will commits the yielded records to the database in batches of given size.
```
alepev, line 58: Important: add (in parenthesis?) that this is decided via the `batch_size` option

alepev (changes requested):
Still some changes, but much better!

Actually there are also couple big points I'm missing in the description (which I failed to bring up in the previous review 🤦‍♀️ ): 
- ~we should mention the `bulk_mode` option in the SQLAlchemy ORM transformations (both normal and batch) as another way to further improve performance - with caveats~ -> EDIT: moved to #126 
- ~a main difference between the different types of transformations, is that Raw SQL and SQLAlchemy queries are ideal for table-to-table transformations, while SQLAlchemy ORM is the way to go if you want to read directly from file or have more complex processing steps, but records insertion will be significantly slower. Basically guiding a bit the user on what to use.. Anything else worth mentioning?~ -> EDIT: moved to #126 

If you like, I could pick up (some of) these points while writing the SQLAlchemy query part :)

alepev:
I also get this sphinx error, but it was there before.. Not sure how to fix it 🤔 
```
/Users/alessia/My_Installs/OHDSI_projects/ohdsi-etl-framework/delphyne/src/delphyne/database/database.py:docstring of src.delphyne.database.database.Database.reflected_metadata:1: WARNING: duplicate object description of src.delphyne.database.database.Database.reflected_metadata, other instance in reference, use :noindex: for one of them
/Users/alessia/My_Installs/OHDSI_projects/ohdsi-etl-framework/delphyne/src/delphyne/database/database.py:docstring of src.delphyne.database.database.Database.schemas:1: WARNING: duplicate object description of src.delphyne.database.database.Database.schemas, other instance in reference, use :noindex: for one of them
```

alepev:
> Still some changes, but much better!
> 
> Actually there are also couple big points I'm missing in the description (which I failed to bring up in the previous review 🤦‍♀️ ):
> 
> * ~we should mention the `bulk_mode` option in the SQLAlchemy ORM transformations (both normal and batch) as another way to further improve performance - with caveats~ -> EDIT: moved to #126
> * ~a main difference between the different types of transformations, is that Raw SQL and SQLAlchemy queries are ideal for table-to-table transformations, while SQLAlchemy ORM is the way to go if you want to read directly from file or have more complex processing steps, but records insertion will be significantly slower. Basically guiding a bit the user on what to use.. Anything else worth mentioning?~ -> EDIT: moved to #126
> 
> If you like, I could pick up (some of) these points while writing the SQLAlchemy query part :)

@MaximMoinat we had a more general issue for guiding the user on performance optimization (#126 ), I moved the two points above there so no need to add this now

docs/transformations.rst
MaximMoinat: For me, it  still renders  fine though!

MaximMoinat: According to these docs, it does not actually need to be the same length (or is it a local delphyne convention? 😝 )

https://docutils.sourceforge.io/docs/user/rst/quickref.html#section-structure

> Titles are underlined (**or over-
> and underlined**) with a printing
> nonalphanumeric 7-bit ASCII
> character.

MaximMoinat: (still, I have updated all titles, does my OCD good...)

docs/transformations.rst
MaximMoinat: @alepev Wait a minute, but the delphyne-template has these execute calls in the `transform()` method... 👀 
https://github.com/thehyve/delphyne-template/blob/master/src/main/python/wrapper.py

MaximMoinat:
Thanks again @alepev for the scrutinous review! Hope it passes the next :-)

docs/transformations.rst
alepev: My bad, indeed it says "at least as long as".. can be longer. Sorry for being so strict 😅

docs/transformations.rst
alepev: True, and in fact in a separate PR we are going to add a documentation page about delphyne-template, to better explain the user where exactly to add these calls. And then we will add a reference here and in other places with a link to the delphyne-template page describing how to implement the wrapper.run() method. I might also move some content around, depending on what looks more clear. For now it's ok like this :)

docs/transformations.rst
MaximMoinat: I just realise that the highlighted comment about over/underlining is not relevant... Can you put the characters on top of the test as well (overlined)?

docs/transformations.rst
MaximMoinat: Do we want to prescribe how to use the `run()` method? I think the only thing that is relevant is that this is the main entry point and the one method to implement when extending the class. Then how you organise your execute calls is totally up to you.

alepev:
review still in progress, but I will complete it today

docs/readme.md
```diff
@@ -0,0 +1,5 @@
+
+### Building the docs
+Edit the .rst, then execute `nox -s docs` in the main projeect folder.
+
+Then open `docs/_build/index.html` in your browser to preview the docs page.
```
alepev, line 5: good idea! but I suggest to add this as a new section in the main readme (root folder).
otherwise this can be mistaken for a source file for generating the docs themselves

docs/transformations.rst
alepev: I also didn't know it was an option 🤔 anyway so far we have used underlined, so I wouldn't change it

alepev (commented):
Well done, it reads smoothly now! Just a suggestion about the readme you added in the last round (see comments above). And then ready to merge

docs/readme.md
alepev: and beware of "projeeeeeeeeeeeect" :P

docs/transformations.rst
```diff
@@ -5,12 +5,118 @@ Transformations
...
+
+
+SQLAlchemy query
+----------------
+Instead of writing plain SQL query, the query can also be written using SQLAlchemy expressions.
+SQLAlchemy translates the expressions into SQL.
+This has the advantage that it can be compiled to any SQL dialect and makes the query agnostic of the used
+Relational Database Management System (RDBMS).
+
+**TBC**
```
alepev, line 122:
```suggestion

.. code-block:: python

    def my_sql_transformation(wrapper):
    
        source_table = wrapper.cdm.<source_table>.__table__
        target_table = wrapper.cdm.<target_table>.__table__

        sel = select([
            source.columns['source_column_1'],
            source.columns['source_column_2'],
            ])\
            .select_from(source)

        ins = target.insert().from_select(sel.columns, sel)
        
        return ins
        
In case the source table is not part of the CDM schema, you can obtain it with the following method, which leverages SQLAlchemy's ability to create reflected table objects from the database itself:

.. code-block:: python
        source_table = wrapper.get_table(schema='my_source_schema', table_name='my_source_table')
        
Inside a wrapper method, the transformations can be called like this, similar to ORM transformations.

.. code-block:: python
    def run(self):
        ...
        self.execute_sql_transformation(my_sql_transformation)

```

alepev:
p.s. yesterday Stefan made some fixes that should resolve the build errors you see above, they probably have nothing to do with your changes (except perhaps for the linter error)

alepev (approved):
OK, I just added a small update from my side (I forgot the links to the wrapper code in the section about SQLAlchemy SQL queries). 
There are some warnings about Database method docstrings but they will disappear when merging to master

Issue #149, opened 2021-02-17
SQL Server testing


Spayralbe:
Add (automated) testing for SQL Server. 
Preferably reusing existing tests, by allowing them to run for different databases.

PR #156, opened 2021-02-23
Fix fixtures

### LINKED ISSUE #154: Better pytest fixtures strategy for improved speed
Look into how to make tests run faster. Especially `test_db` fixture takes up considerable time at the moment since operating at `function` level.

Spayralbe:
Resolves #154 

- Fixes the incorrectly configured fixture dependencies (`@pytest.mark.usefixtures` cannot be used on fixtures but only on tests)
- Make some fixtures available on multiple scopes to allow for more performant test runs when the tests don't require a fresh database for each test

I have the feeling the overall fixture organization could be improved a bit, but it's probably a better idea to revise that as part of #149

### DIFF
--- a/tests/python/conftest.py
+++ b/tests/python/conftest.py
@@ -1,5 +1,6 @@
 import os
 from collections import namedtuple
+from contextlib import contextmanager
 from pathlib import Path
 from typing import Dict
 
@@ -98,13 +99,23 @@ def default_run_config() -> Dict:
 
 
 @pytest.fixture(scope='session')
+def module_scope_db_config() -> Dict:
+    config_file = _CONFIG_DIR / 'module_scope_db_config.yml'
+    return read_yaml_file(config_file)
+
+
+@pytest.fixture(scope='session')
 def default_main_config(default_run_config) -> MainConfig:
     return MainConfig(**default_run_config)
 
 
 @pytest.fixture(scope='session')
-def test_db_uri(default_main_config: MainConfig) -> str:
-    db_config = default_main_config.database
+def module_scope_db_main_config(module_scope_db_config) -> MainConfig:
+    return MainConfig(**module_scope_db_config)
+
+
+def _get_db_string_from_main_config(config: MainConfig) -> str:
+    db_config = config.database
     hostname = db_config.host
     port = db_config.port
     database = db_config.database_name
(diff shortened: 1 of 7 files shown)


alepev:
I just realized there's another way to organize tests: put them in a test class and apply a class fixture. We could make some fixtures class-level, and group tests within the same file (say test_stcm_loader.py) into classes to finely control how many new db instances we need per set of tests. The nice thing is that you still get each test result individually..

Example..

```python

@pytest.fixture(scope="class")
def test_db():
    ...

@pytest.mark.usefixtures("test_db")
class TestStcm1:
    def test_something(self):
        ...

    def test_something_else(self):
        ...

@pytest.mark.usefixtures("test_db")
class TestStcm2:
    """"Another set of tests for STCM that requires a fresh database"""
        ...

```

tests/python/conftest.py
```diff
@@ -113,40 +124,61 @@ def test_db_uri(default_main_config: MainConfig) -> str:
...
     return wrapper_cdm531
 
 
-@pytest.mark.usefixtures("test_db")
 @pytest.fixture(scope='function')
-def cdm600_wrapper_with_tables_created(wrapper_cdm600: Wrapper) -> Wrapper:
+def cdm600_wrapper_with_tables_created(test_db, wrapper_cdm600: Wrapper) -> Wrapper:
     wrapper_cdm600.create_schemas()
     wrapper_cdm600.create_cdm()
     return wrapper_cdm600
```
alepev, line 184: perhaps create a class-level fixture for these, with an additional `wrapper_cdmXXX.drop_cdm()` to make sure no residual data will be left over between test runs?

tests/python/conftest.py
```diff
@@ -113,40 +124,61 @@ def test_db_uri(default_main_config: MainConfig) -> str:
...
+@pytest.fixture(scope='session')
+def test_db_uri(default_main_config: MainConfig) -> str:
+    """Return DB uri for the function-scope DB."""
+    return _get_db_string_from_main_config(default_main_config)
+
+
+@pytest.fixture(scope='session')
+def test_db_module_scope_uri(module_scope_db_main_config: MainConfig) -> str:
+    """Return DB uri for the module-scope DB."""
+    return _get_db_string_from_main_config(module_scope_db_main_config)
```
alepev, line 136: mmm they couldn't use the same config I guess because the two databases will exist at the same time given the different scope? that's inconvenient.. I see now the duplication problem, also with the config

tests/python/model/mapping/test_code_mapper.py
```diff
@@ -1,30 +1,32 @@
...
+def cdm600_wrapper_with_loaded_relationships(test_db_module,
+                                             module_scope_db_main_config: MainConfig
+                                             ) -> Wrapper:
+    wrapper = Wrapper(module_scope_db_main_config, cdm600)
+    wrapper.create_schemas()
+    wrapper.create_cdm()
     wrapper.db.constraint_manager.drop_all_constraints()
     load_concept_relationship(wrapper=wrapper)
     wrapper.db.constraint_manager.add_all_constraints()
     return wrapper
```
alepev, line 26: so here it needs a lot more explicit code indeed.. but it does the trick, the tests for CodeMapper run so much faster now!

PR #160, opened 2021-02-25
A third way of transforming

### LINKED ISSUE #153: Add execute_transformation option for SQLAlchemy Query
Currently there are options for executing queries via the ORM, or by using raw SQL queries. It would be useful to have a third option using the SQLAlchemy Query definition. That way you will have the same performance as raw SQL, but you can still make the queries dialect-agnostic.

- [x] Add the execution method to wrapper
- [x] Add collection method for transformation metadata

alepev:
Fixes #153 

Sample transformation related to this in https://github.com/thehyve/delphyne-template/pull/67

### DIFF
--- a/src/delphyne/model/raw_sql_wrapper.py
+++ b/src/delphyne/model/raw_sql_wrapper.py
@@ -11,6 +11,7 @@ from sqlalchemy.engine.result import ResultProxy
 
 from .etl_stats import EtlTransformation, open_transformation
 from .._paths import SQL_TRANSFORMATIONS_DIR
+from ..cdm.schema_placeholders import CDM_SCHEMA
 from ..config.models import MainConfig
 from ..database.database import Database
 from ..util.helper import replace_substrings
@@ -119,8 +120,9 @@ class RawSqlWrapper:
                 as (session, transformation_metadata):
             query = statement(self)
             result = session.execute(query)
-            status_message = result.context.cursor.statusmessage
-            self._collect_query_statistics(result, status_message, transformation_metadata)
+            query_string = str(query).replace(CDM_SCHEMA,
+                                              self.db.schema_translate_map[CDM_SCHEMA])
+            self._collect_query_statistics(result, query_string, transformation_metadata)
 
     @staticmethod
     def apply_sql_parameters(parameterized_query: str, sql_parameters: Dict[str, str]) -> str:
--- a/src/delphyne/model/raw_sql_wrapper.py
+++ b/src/delphyne/model/raw_sql_wrapper.py
@@ -6,7 +6,7 @@ from collections import Counter
 from pathlib import Path
 from typing import Callable, Dict, Union, Optional
 
-from sqlalchemy import text
+from sqlalchemy import text, Table
 from sqlalchemy.engine.result import ResultProxy
 
 from .etl_stats import EtlTransformation, open_transformation
@@ -123,6 +123,27 @@ class RawSqlWrapper:
             query_string = str(query).replace(CDM_SCHEMA,
                                               self.db.schema_translate_map[CDM_SCHEMA])
             self._collect_query_statistics(result, query_string, transformation_metadata)
+
+    def get_cdm_table(self, table_name: str) -> Table:
+        """
+        Get a SQLAlchemy Table object from the cdm schema.
+
+        Parameters
+        ----------
+        table_name : str
+            Name of the table to retrieve. An error message will be
+            displayed if the table can't be found.
+
+        Returns
+        -------
+        Table
+        """
+        cdm_schema = self.db.schema_translate_map[CDM_SCHEMA]
+        full_table_name = f'{CDM_SCHEMA}.{table_name}'
+        target_table = self.db.base.metadata.tables.get(full_table_name, None)
+        if target_table is None:
+            logger.error(f'Table with name {table_name} not found in {cdm_schema} schema')
+        return target_table
 
     @staticmethod
     def apply_sql_parameters(parameterized_query: str, sql_parameters: Dict[str, str]) -> str:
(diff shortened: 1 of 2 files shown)


alepev:
Still have to figure the details in the template repo.. the example we have in the mapper are from schema to schema in the same database, this should be from file to database.. unless we agree it's not a good use case for that. maybe we should actively suggest to only use this new method for schema to schema transformations?

src/delphyne/model/raw_sql_wrapper.py
```diff
@@ -90,13 +90,38 @@ def execute_sql_query(self, query: str, query_name: str) -> None:
...
+        Returns
+        -------
+        None
+        """
+        logger.info(f'Executing transformation: {statement.__name__}')
+        with self.db.tracked_session_scope(name=statement.__name__, raise_on_error=False) \
+                as (session, transformation_metadata):
+            query = statement(self)
+            result = session.execute(query)
+            status_message = result.context.cursor.statusmessage
```
alepev, line 122: it was very hard to find this, perhaps there's a better place to extract the query string from..

src/delphyne/model/raw_sql_wrapper.py
```diff
@@ -118,15 +143,17 @@ def apply_sql_parameters(parameterized_query: str, sql_parameters: Dict[str, str
...
+                                  result: ResultProxy,
+                                  query: str,
+                                  transformation_metadata: EtlTransformation
+                                  ) -> None:
         query_type = self._parse_query_type(query)
         target_table: str = self._parse_target_table_from_query(query)
         row_count: int = result.rowcount
 
+        logger.info(f'Saving {row_count} objects')
+
```
alepev, line 156: not sure if it's really worth adding this..

src/delphyne/model/raw_sql_wrapper.py
alepev: Replaced with a more complete source for the query string

src/delphyne/model/raw_sql_wrapper.py
```diff
@@ -90,13 +91,60 @@ def execute_sql_query(self, query: str, query_name: str) -> None:
...
+        Returns
+        -------
+        Table
+        """
+        cdm_schema = self.db.schema_translate_map[CDM_SCHEMA]
+        full_table_name = f'{CDM_SCHEMA}.{table_name}'
+        target_table = self.db.base.metadata.tables.get(full_table_name, None)
+        if target_table is None:
+            logger.error(f'Table with name {table_name} not found in {cdm_schema} schema')
+        return target_table
```
alepev, line 146: I might want to change this into raising an exception.. at the moment the ETL execution goes on. What would be a suitable exception? 🤔

Spayralbe (commented):
✨ I think the flexibility this style of transformation provides is really nice!

src/delphyne/model/raw_sql_wrapper.py
```diff
@@ -90,13 +91,60 @@ def execute_sql_query(self, query: str, query_name: str) -> None:
...
+        -------
+        None
+        """
+        logger.info(f'Executing transformation: {statement.__name__}')
+        with self.db.tracked_session_scope(name=statement.__name__, raise_on_error=False) \
+                as (session, transformation_metadata):
+            query = statement(self)
+            result = session.execute(query)
+            query_string = str(query).replace(CDM_SCHEMA,
+                                              self.db.schema_translate_map[CDM_SCHEMA])
```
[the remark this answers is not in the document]
alepev: I just wish I could find these things as easily.. literally spent hours to figure where to get the information from 😩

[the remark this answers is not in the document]
Spayralbe: The debugger is your friend! :D

src/delphyne/model/raw_sql_wrapper.py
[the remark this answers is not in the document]
alepev: Does delphyne already support custom schema names? I seem to remember it does.. Now I'm wondering if I could have made things easier in the mapper repo too :/ I'll add a schema to do some tests with the template locally.. For cdm tables, something like `Person.__table__` also works indeed

src/delphyne/model/raw_sql_wrapper.py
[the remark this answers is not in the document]
alepev: WHY do I even bother to try with SQLAlchemy documentation??? 🤦‍♀️

src/delphyne/model/raw_sql_wrapper.py
alepev: Updated

src/delphyne/model/raw_sql_wrapper.py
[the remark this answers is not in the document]
alepev: Changed the method as suggested, I actually had something very similar in the mapper repo. Except there the source schema is fixed, so it doesn't need to be passed to the function.. You can test it in the template script by replacing one of the `<Table>.__table__` statements, seems to work :)

src/delphyne/model/raw_sql_wrapper.py
```diff
@@ -90,13 +91,71 @@ def execute_sql_query(self, query: str, query_name: str) -> None:
...
+        The table object is obtained through metadata reflection; useful
+        when retrieving tables that are not defined in the ORM model
+        (e.g. table is in a custom source schema). For tables that
+        already have an ORM definition, this method is equivalent to
+        using <TableName>.__table__.
+
+        Parameters
+        ----------
+        schema: str
+            Name of the schema to retrieve the table from.
```
Spayralbe, line 141: Something to add:

> This may also be a schema placeholder, in which case it will be translated to the runtime schema name.

alepev: Good point, even more clear for the user

Spayralbe (approved):
🦭

alepev:
> 🦭

I'm not sure what that thing is, but glad I can merge :D

PR #162, opened 2021-03-03
Stcm version cleanup

### LINKED ISSUE #161: Cleanup old source_vocabulary_ids in STCM_version
If I change the vocabulary ids in `stcm_versions.tsv` (say the custom vocabularies I was referencing have been deleted and I want to use different ones), the old vocabularies are not automatically removed from the `SourceToConceptMapVersion` table in the database. Once the constraints are re-applied, I get a FK constraints error because the deleted vocabularies are still being referenced. 

Solution: implement stcm version cleanup.

alepev:
Adds check that deletes stcm records from the database if they are not present in the current stcm and stcm_version files.

Fixes #161

### DIFF
--- a/tests/python/model/stcm/test_stcm_loader.py
+++ b/tests/python/model/stcm/test_stcm_loader.py
@@ -126,7 +126,8 @@ def test_load_stcm(cdm600_wrapper_no_constraints: Wrapper, base_stcm_dir: Path, 
         versions = get_all_stcm_versions(wrapper)
         assert records == [('code3', 'MY_VOCAB2')]
         assert versions == [('MY_VOCAB2', '0.2')]
-    assert "Deleting obsolete stcm and stcm_version records" in caplog.text
+    assert "Deleting obsolete stcm and stcm_version records for vocabulary_ids:" \
+           " {'MY_VOCAB1', 'MY_VOCAB2'}" in caplog.text
 
 
 def test_stcm_delete(cdm600_wrapper_no_constraints: Wrapper, base_stcm_dir: Path):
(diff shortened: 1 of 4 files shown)


alepev:
I shall add some tests 🤔 but current ones are still passing

tests/python/model/stcm/test_stcm_loader.py
```diff
@@ -105,6 +119,15 @@ def test_load_stcm(cdm600_wrapper_no_constraints: Wrapper, base_stcm_dir: Path,
...
     assert "No new STCM versions provided" in caplog.text
 
+    # MY_VOCAB2 updated, MY_VOCAB1 removed
+    with mock_stcm_paths(base_stcm_dir, 'stcm3'), caplog.at_level(logging.INFO):
+        wrapper.vocab_manager.stcm.load()
+        records = get_all_stcm_records(wrapper)
+        versions = get_all_stcm_versions(wrapper)
+    assert records == [('code3', 'MY_VOCAB2')]
+    assert versions == [('MY_VOCAB2', '0.2')]
+    assert "Deleting STCM and STCM version records for vocabulary_ids: MY_VOCAB1" in caplog.text
```
alepev, line 129: Help! This works when playing with STCM in the template, but I don't get the expected result here.. Somehow the records are not deleted, am I missing something here?

alepev:
I found the logic fault that was causing the failed tests only with repeated `StcmLoader` calls, but not when using the template 🎉

alepev:
On a side note: what's wrong with GitHub??? It's grouping together commits from last week with commits I just pushed, those were not 8 days ago!
Screenshot 2021-03-11 at 13 20 04

Spayralbe (approved):
👏

alepev:
Yay!

PR #163, opened 2021-03-04
Document STCM Loader

### LINKED ISSUE #123: Document STCM usage
Depends on https://github.com/thehyve/omop-etl-wrapper/issues/41

alepev:
Fixes #123 

In the end for the structure I went like this:
- STCM loading in its own section, as for standard & custom vocabularies
- STCM mappings (which is actually a separate lookup provided by the Wrapper, not StcmLoader) in the Semantic mapping tools

Also cleaned up some things along the way.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,6 +13,7 @@ copyright = f"2021, {author}"
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.napoleon",
+    "sphinx.ext.autosectionlabel",
     # "numpydoc",
 ]
 
--- a/docs/semantic_mapping.rst
+++ b/docs/semantic_mapping.rst
@@ -14,13 +14,13 @@ Delphyne's :class:`.CodeMapper` class enables the creation of mapping dictionari
 terms to valid standard concept_ids. Once created, a mapping dictionary can be used in any transformation to quickly
 lookup mappings for any source term.
 
-Mappings are built on information extracted from the CONCEPT and CONCEPT_RELATIONSHIP tables.
-For example, it is possible to automatically map ICD10CM ontology terms to their SNOMED standard concept_id equivalent;
-the only prerequisite is to have both vocabularies loaded to your database (including CONCEPT_RELATIONSHIP information).
+Mappings are built on information extracted from the CONCEPT and CONCEPT_RELATIONSHIP tables. For example, it is
+possible to automatically extract mappings from ICD10CM ontology terms to their equivalent standard concept_id
+(typically SNOMED).
 
 Note that creating a mapping dictionary with this method is only possible for **official OMOP vocabularies**;
 if a source vocabulary is not available from `Athena <https://athena.ohdsi.org/vocabulary/list>`_,
-other mapping methods should be considered.
+alternative methods should be considered.
 
 Creating a mapping dictionary
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/semantic_mapping.rst
+++ b/docs/semantic_mapping.rst
@@ -5,12 +5,12 @@ Semantic mapping tools
     :local:
     :backlinks: none
 
-The section describes available Delphyne's tools to map source data to standard OMOP concept_ids during ETL execution.
+The section describes available delphyne's tools to map source data to standard OMOP concept_ids during ETL execution.
 
 Vocabulary-based mappings
 -------------------------
 
-Delphyne's :class:`.CodeMapper` class enables the creation of mapping dictionaries from non-standard OMOP vocabulary
+delphyne's :class:`.CodeMapper` class enables the creation of mapping dictionaries from non-standard OMOP vocabulary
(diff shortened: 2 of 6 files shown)


docs/conf.py
```diff
@@ -13,6 +13,7 @@
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.napoleon",
+    "sphinx.ext.autosectionlabel",
```
alepev, line 16: to enable `:ref:` directives

docs/semantic_mapping.rst
```diff
@@ -78,3 +79,21 @@ a single :class:`.CodeMapping` object with both ``source_concept_id`` and ``targ
...
+Looking up STCM mappings
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+You can lookup mapping information for a single source code at a time as follows:
+
+.. code-block:: python
+
+   mapping = wrapper.lookup_stcm(source_vocabulary_id='MY_VOCAB', source_code='ABC')
+
+The result is a single standard OMOP concept_id, or ``0`` if nothing is found.
```
alepev, line 99: I just realized this does not support mappings of the same source code to multiple target concept_ids, even if in principle the STCM table does!

docs/stcm.rst
```diff
@@ -0,0 +1,133 @@
...
+in the STCM format (*File > Export source_to_concept_map*).
+
+STCM files may contain source codes from one or more custom vocabularies.
+When a STCM file contains records from a single custom vocabulary,
+it's good practice to prepend the vocabulary_id to the file name (e.g. MYVOCAB_stcm.csv).
+This way, if the mapping version associated with that vocabulary hasn't changed (see `Versioning`_),
+the file will be ignored without needing to parse the file contents.
+
+Below is an **STCM file example**; in addition to the fields shown, it should contain
+``valid_start_date``, ``valid_end_date`` (mandatory), and ``source_code_description``, ``invalid_reason`` (optional):
```
alepev, line 26: I thought of providing a tabular example because you cannot just peek at your fresh Athena vocabulary download to see what columns to put in there.. not sure though if it is too confusing like this (with some of the fields omitted). If I put all the columns in, it looks quite ugly 😅

docs/semantic_mapping.rst
alepev: It wouldn't actually be too much work to make the CodeMapper look into the STCM table for custom mappings, and provide them through the same interface.. New issue? 😬

Spayralbe: Ah good you're pointing this out. I did not realize that could be a valid return value.
Shall we first create an issue about supporting the return of multiple values? Then we can discuss later whether we want to move the STCM lookup to the `CodeMapper`.

alepev: Yes integrating would be a big change, let's start with baby steps

docs/stcm.rst
Spayralbe: Nah I think it's nice to have an example :)

Spayralbe (commented):
Great work! I have nothing to complain about for these additions 😄

One thing that would be good to add for both the custom vocab files and STCM files, is some information on file headers. I.e. always a header should be present in these files, of which the field names should match the column names of the database tables they map to (in lowercase).

docs/semantic_mapping.rst
Spayralbe: Added #173

Spayralbe:
Ooh and shall we stick to the lowercase name "delphyne" instead of "Delphyne"? 
As that is what we now already use for the github name and on PyPI.

alepev:
> Great work! I have nothing to complain about for these additions 😄
> 
> One thing that would be good to add for both the custom vocab files and STCM files, is some information on file headers. I.e. always a header should be present in these files, of which the field names should match the column names of the database tables they map to (in lowercase).

I will add something to both sections :)

alepev:
OK fixed both things, let me know if you want to take another look @Spayralbe

docs/stcm.rst
```diff
@@ -107,12 +110,14 @@ and replace the previous call with the following in the Wrapper's run method:
...
 Versioning
 ----------
+
 STCM records are always associated with a mapping version (``stcm_version``) through their ``source_vocabulary_id``;
 this relationship is captured in the SOURCE_TO_CONCEPT_MAP_VERSION table.
 
 When you update the mapping version for a given ``source_vocabulary_id`` in the ``stcm_versions.tsv`` file,
 this will update the SOURCE_TO_CONCEPT_MAP_VERSION table, and cause all SOURCE_TO_CONCEPT_MAP records associated with
 that vocabulary to be dropped and replaced with new records from the provided STCM files, if any.
+If you completely remove a ``source_vocabulary_id`` from ``stcm_versions.tsv``, associated records will also be dropped.
```
Spayralbe, line 120:
```suggestion
If you remove a ``source_vocabulary_id`` from ``stcm_versions.tsv``, associated records will also be dropped.
```

Spayralbe (approved):

PR #167, opened 2021-03-11
Update config options

### LINKED ISSUE #166: Rename run options
**How would this feature be useful?**
The current run options with `skip_` cause a double negative which is very confusing.
(and yes, I was myself guilty of introducing the `skip_` syntax in the first place...)

**Describe the solution you'd like**
Rename
```yaml
run_options:
  skip_vocabulary_loading: False
  skip_custom_vocabulary_loading: False
  skip_source_to_concept_map_loading: False
```
to
```yaml
run_options:
  do_load_standard_vocabulary: True
  do_load_custom_vocabulary: True
  do_load_source_to_concept_map: True
```

alepev:
Fixes #166 

Related to https://github.com/thehyve/delphyne-template/pull/68

### DIFF
--- a/docs/custom_vocab.rst
+++ b/docs/custom_vocab.rst
@@ -45,7 +45,7 @@ Files have to be placed in te following folder:
 Add to pipeline
 ---------------
 
-Make sure the ``skip_custom_vocabulary_loading`` option in your config.yml is set to False.
+Make sure the ``load_custom_vocabulary`` option in your config.yml is set to ``True``.
 
 If you're also using STCM files, the recommended way to load
 the files is with the following call as part of your Wrapper's run method:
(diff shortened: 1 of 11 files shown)


alepev:
Mmm not sure I needed to change the internals to this extent, but I thought it would make things clearer also when refactoring the code 🤔

Spayralbe (commented):
LGTM. There are however also a few references to these config options in the docs that will have to be updated.

alepev:
> LGTM. There are however also a few references to these config options in the docs that will have to be updated.

Indeed! I wondered if I should make edits here and risk conflicts with the STCM docs PR.. I'll first merge that and then edit the relevant parts. But I should have written it somewhere, I almost forgot! 🙈

PR #187, opened 2021-04-13
Remove separate database_exists check

Spayralbe:
Since SQLALchemy-Utils version 0.37 there is no a longer a distinction between invalid connections and non-existing databases when calling `database_exists`. Therefore the connection testing logic is now simplified a bit and the logger will just show the original error in case a database connection could not be established.

Makes SQLALchemy-Utils a test-only dependency.

### DIFF
--- a/src/delphyne/database/constraints/constraint_manager.py
+++ b/src/delphyne/database/constraints/constraint_manager.py
@@ -72,11 +72,14 @@ class _DbCheckConstraints:
     # these for the dialect in use.
     def __init__(self, database: Database):
         self._db = database
-        self.chk_support = self._check_reflection_support()
-
         self._dialect_method_lookup = {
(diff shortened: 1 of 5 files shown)


src/delphyne/database/constraints/constraint_manager.py
```diff
@@ -72,12 +72,15 @@ class _DbCheckConstraints:
...
-
         self._dialect_method_lookup = {
             'mssql': self._get_chk_constraints_mssql
         }
 
+    @property
+    @lru_cache()
+    def chk_support(self) -> bool:
+        return self._check_reflection_support()
+
```
Spayralbe, line 83: This is actually a bit unrelated, but with these changes, it will only check for check constraint support after the database connection has been tested.

alepev (approved):
Looks good! Minus one dependency, yay!

PR #188, opened 2021-04-14
Support multiple STCM targets

### LINKED ISSUE #173: Support multiple targets for STCM lookup
Currently the `lookup_stcm` method expects a `one_or_none` query result, whereas it is valid to have multiple target concept ids for one source code.

The lookup method should therefore be updated to allow multiple target concept id values.

Spayralbe:
Fixes #173 

- Return STCM lookup results as a tuple, that can contain zero, one or multiple target_concept_id values
- Move the lookup method to the `CodeMapper` class
- Update STCM lookup docs

### DIFF
--- a/docs/semantic_mapping.rst
+++ b/docs/semantic_mapping.rst
@@ -94,6 +94,10 @@ You can lookup mapping information for a single source code at a time as follows
 
 .. code-block:: python
 
-   mapping = wrapper.lookup_stcm(source_vocabulary_id='MY_VOCAB', source_code='ABC')
+   mapping = wrapper.code_mapper.lookup_stcm(source_vocabulary_id='MY_VOCAB', source_code='ABC')
 
-The result is a single standard OMOP concept_id, or ``0`` if nothing is found.
+The result is a tuple containing one or multiple target_concept_id values if available.
+If no target_concept_id could be found for the provided source_code, an empty tuple is returned.
+
+.. note::
+   STCM lookups are cached, meaning that recurrent lookups will not require a database query.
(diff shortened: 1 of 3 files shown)


alepev (commented):
`semantic_mapping.rst` still links to the original Wrapper method, update this section: 

```
STCM mappings
-------------
The :class:`.Wrapper` class provides a :meth:`~.Wrapper.lookup_stcm()` method to extract mappings from the
SOURCE_TO_CONCEPT_MAP table. Note that you will need to populate the table yourself before being able to use this
method (see :ref:`stcm:Source to concept map` for instructions).
```
Also, shall we add some tests? Or shall we leave that for the next big refactoring?

Spayralbe:
Thanks! Completely overlooked those references.

alepev (approved):
Then you are good to go!

Spayralbe:
@alepev Ooh and about the tests: Yes I'd say we keep that for after the implementation of the new mapping table.

alepev:
> @alepev Ooh and about the tests: Yes I'd say we keep that for after the implementation of the new mapping table.

Yes I also just thought that it's not an efficient use of time for now.. Can you add a separate issue for the tests pointing to the CodeMapper method you just added? We can say that it has to happen after refactoring. 

..Actually, we don't rven have a specific issue for the refactoring itself 🤔  Only something mentioned in #39 in relation to the VariableConceptMapper, which I'm not even sure we'll pick up once we can load unit and value mappings directly to the STCM 🤔  

too much 🤔 🤔 🤔

Spayralbe:
https://i.giphy.com/media/xUPGcz2H1TXdCz4suY/giphy.webp

alepev:
That's precisely how I spend most of my days

Issue #124, opened 2020-11-03
Document config files options


alepev:
Document use cases / give examples of how to customize:
- [x] general config
- [x] source config
- [x] log config

PR #192, opened 2021-05-03
Add source data docs

### LINKED ISSUE #122: Document how to work with source data files
Document `SourceData` and `SourceFile` class methods with use cases for different memory / performance limits scenarios.

Spayralbe:
Closes #122
Depends on #124 

Add documentation for handling source data.

TODO:
- [x] Resolve the two references to not yet available config documentation
- [x] Add links for references to source_config and config


docs/source_data.rst
```diff
@@ -0,0 +1,128 @@
...
+
+    source = wrapper.source_data.get_source_file('source_table.csv')
+
+The resulting object can then be used to access the file contents in various ways, depending on the file type.
+
+Delimited text files
+--------------------
+Delimited text files (e.g. csv) can be used in transformations in several ways.
+Any file-specific information (e.g. encoding, delimiter) needed to read the file should be provided in the
+source_config.
```
alepev, line 33: In other places I refer to files using the full extension (e.g. `source_config.yml`), and also the highlighting with ``. I'm not sure about the latter (perhaps better use it only for code), but it helps to distinguish files and folders names in the text.. Shall we agree on a convention? Or can it be that we did already, but I forgot? 😅

docs/source_data.rst
```diff
@@ -0,0 +1,128 @@
...
+-----------------------
+Source data files can be read directly to be used in :ref:`transformations:SQLAlchemy ORM` style transformations.
+This requires a `source data path <TODO>`_ to be set in the main config file,
+and the presence of a `source_config <TODO>`_.
+
+To use a file in a transformation, first define the source by providing the file name:
+
+.. code-block:: python
+
+    source = wrapper.source_data.get_source_file('source_table.csv')
```
alepev, line 25: Would it be a good idea to add links to the API references for `SourceData` and `SourceFile` classes in case one wants to jump directly to those? Personally I like it.. it does mean more documentation references updates in case we ever change the names, but it seems unlikely

docs/source_data.rst
```diff
@@ -0,0 +1,128 @@
...
+    df = source.get_csv_as_df(apply_dtypes=True, cache=False)
+    # Do some preprocessing on df here
+    source.cache_df(df)
+
+The next time :meth:`~.SourceFile.get_csv_as_df()` is called for this source, the cached dataframe will be returned,
+meaning the preprocessing is already performed.
```
alepev, line 104: You will still need to repeat the initial `source = wrapper.source_data.get_source_file('source_table.csv')` to obtain a SourceFile object, right? And then it is matched to the available cached object based on the name..?

docs/source_data.rst
```diff
@@ -0,0 +1,128 @@
...
+.. code-block:: python
+
+    source = wrapper.source_data.get_source_file('source_table.sas7bdat')
+    df = source.get_sas_as_df(apply_dtypes=False)
+    # Do some preprocessing on df here
+    source.cache_df(df)
+
+Read from database table
+------------------------
+If using :ref:`raw sql queries <transformations:Raw SQL>`
```
alepev, line 124: SQL uppercase

docs/source_data.rst
```diff
@@ -0,0 +1,128 @@
...
+or the :ref:`SQL expression language <transformations:SQLAlchemy query>`
+for transformations, the source data will first need to be inserted into the database.
```
alepev, line 126: It's funny, I was thinking of it more the other way round.. "if you have your data in a database, then you should these transformations" :P I thought of it more of a use case for when people already come to us with a database, but perhaps it does make sense to suggest to load the data as the first step (assuming it is the final version of the data), to be able to use those more performant transformation styles. (Just rambling, you can leave it as is)

alepev (commented):
Some suggestions but basically spot on. 

I was wondering if it's also worth mentioning `get_line_count()`, which is mainly for the ETL stats report.. could be nice for the user, but I'm fine with skipping.

docs/source_data.rst
Spayralbe: Exactly

docs/source_data.rst
```diff
@@ -0,0 +1,128 @@
...
+The same caching options apply as for delimited text files.
+
+.. code-block:: python
+
+    source = wrapper.source_data.get_source_file('source_table.sas7bdat')
+    df = source.get_sas_as_df(apply_dtypes=False)
+    # Do some preprocessing on df here
+    source.cache_df(df)
+
+Read from database table
```
alepev, line 122: Last minute comment:
we have some examples for how to do this in the Transformations section,
- `source_table = wrapper.cdm.<source_table>.__table__` if the table is part of the model,
- `source_table = wrapper.get_table(schema='my_source_schema', table_name='my_source_table')` if not

Would they better belong here? Or leave them there, but explain that options to read from database are explained in detail in that section

docs/source_data.rst
Spayralbe: You mean inside this code block (not sure how to do that)? Or just to reference them somewhere in the text?

alepev: In the text

docs/source_data.rst
Spayralbe: Yes, in principle you can use all three transformation options regardless of whether you have the data in a database or as delimited text files. It just means you might have to export/import first to be able to use them.
I will expand this part a little bit.

docs/source_data.rst
Spayralbe: Maybe we can make them reference the relevant header (as soon as it exists) in the config docs? Then they will highlighted as a link.

docs/source_data.rst
Spayralbe: I think it's ok to refer to the transformations section from here.

docs/source_data.rst
Spayralbe: Reference added

alepev (approved):
Ready to merge? 😬

Issue #74, opened 2020-12-15
Split vocab transformation from clinical CDM in EtlStats


Spayralbe:
Currently the `EtlStats` summary report that is logged at the very end, does not distinguish between `EtlTransformation` objects that only affect vocabulary tables and those that affect the other CDM tables.

DoD: Create a separate summary report for the vocabulary table transformations.

PR #193, opened 2021-05-17
Add configuration docs

### LINKED ISSUE #124: Document config files options

Spayralbe:
Add configuration files documentation.

Closes #124


docs/config.rst
```diff
@@ -0,0 +1,273 @@
...
+
+The database config values are used to establish a connection to the database.
+The contents are highly similar to SQLAlchemy's
+`URL variables <https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.engine.URL>`_.
+
+.. _drivername-anchor:
+
+**drivername**
+
+Database backend and (optionally) driver name, e.g. ``mssql+pyodbc``.
```
alepev, line 29: I was wondering, are there general conventions around the names of backends & drivers? would it make sense to list the ones we support?

docs/config.rst
```diff
@@ -0,0 +1,273 @@
...
+
+**query**
+
+Any additional options to be passed on to the DBAPI (optional).
+May consist of multiple key-value pairs, e.g.
+
+.. code-block:: yaml
+
+    query:
+        driver: 'ODBC Driver 17 for SQL Server'
```
alepev, line 72: is there any specific query required for compatibility with certain DBMSs? I vaguely remember something with MySQL Server, was it the one you wrote in this example?

docs/config.rst
```diff
@@ -0,0 +1,273 @@
...
+
+The (relative) path to the source data folder.
+Only needed when using source data files.
+
+Schema translate map
+^^^^^^^^^^^^^^^^^^^^
+
+The ``schema_translate_map`` is used to replace placeholder schema names in your cdm model with the actual names
+at runtime. It should at least contain the schema names of the target OMOP CDM schema (``cdm_schema``)
+and the ``vocabulary_schema`` (can be the same as ``cdm_schema``).
```
alepev, line 89: just occurred to me: what happens when you use `drop_cdm()` in the wrapper if cdm and vocab schemas are the same? would they all be dropped or all be skipped? perhaps we should put a big warning about this somewhere 😬

docs/config.rst
```diff
@@ -0,0 +1,273 @@
...
+
+If False, all contents in ``resources/vocabularies/custom`` will be ignored,
+regardless of whether you call load_custom_vocabularies in your code.
+
+.. _config-run_options-load_source_to_concept_map-anchor:
+
+**load_source_to_concept_map (boolean)**
+
+If False, all contents in ``resources/vocabularies/source_to_concept_map``
+will be ignored, regardless of whether you call load_stcm in your code.
```
alepev, line 127: should the calls to the methods be made into links to those methods, or at least highlighted? which reminds me, at the moment we have the same style for file paths and inline code 🤔

docs/config.rst
```diff
@@ -0,0 +1,273 @@
...
+**write_reports (boolean)**
+
+Write two additional tsv files in the logs folder, with more
+detailed information on source data counts and ETL transformations.
+
+SQL parameters
+^^^^^^^^^^^^^^
+
+``sql_parameters`` is an optional key-value map. It can be used together with parameterized
+raw SQL queries, of which the placeholders (e.g. schema names, values)
```
alepev, line 140: also here, I would suggest to link to the Raw SQL transformations docs

docs/config.rst
```diff
@@ -0,0 +1,273 @@
...
+
+Row count
+^^^^^^^^^
+
+.. _source_config-count_source_rows-anchor:
+
+**count_source_rows (boolean)**
+
+If True, for each source file count the number of rows and report these.
+The row count will only be determined for files that are not in binary format.
```
alepev, line 191: this will only be saved in the additional log file written when `write_reports` is True, or are those two independent? can you log this without saving it to log file, and save it to log file without logging it?

docs/config.rst
```diff
@@ -0,0 +1,273 @@
...
+
+**source_data_folder**
+
+The (relative) path to the source data folder.
+Only needed when using source data files.
+
+Schema translate map
+^^^^^^^^^^^^^^^^^^^^
+
+The ``schema_translate_map`` is used to replace placeholder schema names in your cdm model with the actual names
```
alepev, line 87:
```suggestion
The ``schema_translate_map`` is used to replace placeholder schema names in your CDM model with the actual names
```

docs/config.rst
```diff
@@ -0,0 +1,273 @@
...
+          column_b: Int64
+          column_c: datetime64[ns]
+          column_d: float64
+
+
+Logging configuration
+---------------------
+
+The logging config is a recommended, yet optional configuration file that can be used to control the logging output.
+If using, it is expected at ``config/logging.yml``.
```
alepev, line 269:
```suggestion
If used, it is expected at ``config/logging.yml``.
```

docs/config.rst
Spayralbe: Anything that is accepted by SQLAlchemy will work here as well

docs/config.rst
Spayralbe: I thought I tested this already, but trying it again now, it turns out the opposite happens: nothing is dropped.
Time to add some more tests 😬

alepev (commented):
Clear and nicely structured 🏗️ 

General comment on top of the small ones: you mention the sample file for logging config but not the other two configs, I suggest to point to the sample file for all of them, or none (also an option, since mentioned in _Getting started with delphyne_).

For the main config, no folder location is mentioned because indeed delphyne doesn't expect a specific file path, but I was wondering if we should say this explicitly.

alepev:
Oh, and I forgot! I think this should be also linked to a PR in delphyne-template to remove the descriptions from the config sample files :P

docs/config.rst
Spayralbe: See #196

docs/config.rst
Spayralbe: References added. Will wait with modifying the paths markup, until we made a decision on #197

docs/config.rst
Spayralbe: Agreed. Added the reference.

docs/config.rst
Spayralbe: Clarification added.

Spayralbe:
> Oh, and I forgot! I think this should be also linked to a PR in delphyne-template to remove the descriptions from the config sample files :P

That's a very good point😄 Now added at https://github.com/thehyve/delphyne-template/pull/74

alepev (approved):
Heel goed! 💯

PR #194, opened 2021-05-20
Transformation docs updates

Spayralbe:
Add info on
- Session injection in regular transformations
- Running transformations in bulk mode


docs/transformations.rst
```diff
@@ -56,6 +61,21 @@ All returned records are committed to the database at the same time.
         return records_to_insert
 
 While this is straightforward, it does require to store all output records in memory.
+
+Session injection
+~~~~~~~~~~~~~~~~~
```
alepev, line 66: Uhm I did not realize `"` comes before `~`.. in _Getting started with delphyne_ I went straight from `^` to `~` 🤔 
Either way, at this level of sub-section the title doesn't seem to be displayed in any particular format.. I wanted it to be highlighted somehow, so instead I made it bold (see _8.1. Write transformations in Python_ below)  https://delphyne.readthedocs.io/en/latest/delphyne_template.html#write-source-data-transformations
maybe I shouldn't have, the section level info is lost that way.. 🙄 

I'm curious to see if once published on Read the docs, _Session injection_ is actually going to look different, if so I will also change the style of those sub-sub-sub sections in delphyne_template.rst.

alepev: Although if you DON'T want to have this title as part of the ToC at the top, making it bold is actually not a bad idea :P

Spayralbe: Turns out you can use bold notation for a header as well :p
Looks better indeed

PR #196, opened 2021-05-20
drop_cdm

### LINKED ISSUE #195: Drop cdm tables with single schema
Calling `drop_cdm` when using a single schema for both the vocab and non-vocab tables, results in none of the tables being dropped. It should still drop the tables that do not use the vocabulary placeholder schema name.

Spayralbe:
Fixes #195 

- Fix `drop_cdm` when using a single schema
- Add tests for `drop_cdm`
- Move lint dependencies to `extras_require` of setup.py


alepev (approved):
It perfectly makes sense, now that my brain allowed me to process it 💯

PR #205, opened 2021-06-18
Config docs escape character

Spayralbe:
Related to https://github.com/thehyve/delphyne-template/pull/77

- Add info on escape characters in YAML config values
- Fix broken links in docs


alepev (approved):
Thanks for the additional fixes! :)

